PR #28, opened 2017-03-19
Fixes #6 Added user setting for participants and votes reset

### LINKED ISSUE #6: Adding extra dates without reset
It would be nice to be able to add extra dates, without resetting everyone's votes.

nerdsRob:
Enable users to choose whether they want to reset all participants and their votes after editing their event's date options. Default setting is defined as `false`.

### DIFF
--- a/api/event/event.controller.js
+++ b/api/event/event.controller.js
@@ -75,6 +75,10 @@ exports.update = function(req, res){
                     communicator.emit('event:update:creator.email', updatedEvent, event);
             }
 
+            if (event.creator.shouldResetParticipants != updatedEvent.creator.shouldResetParticipants) {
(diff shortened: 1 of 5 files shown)


lukevella:
Unfortunately unless you are adding days that are later than the current options this will not work.

Say you want to remove the first date on the list:
screen shot 2017-03-20 at 9 05 50 pm

Notice how the votes from the 21 Mar transfer to 24 Mar.
screen shot 2017-03-20 at 9 06 06 pm

To fix this, would require a rework of how the votes are stored in the database. Admittedly I did a very poor job on this the first time round.

nerdsRob:
Thanks for pointing out this use case and the problems w/ the current model. I will give it another shot, since I really would like to use this feature myself 🤓

PR #64, opened 2018-12-03
Import Google Fonts using https

dhayab:
Currently, remote fonts from Google are loaded using an `@import` rule. Because the URL is not secured, browser blocks the request and ask for explicit permission from the user to enable mixed content.

> Mixed Content: The page at 'https://rallly.co/' was loaded over HTTPS, but requested an insecure stylesheet 'http://fonts.googleapis.com/css?family=Roboto+Slab:700|Roboto+Condensed:400,700'. This request has been blocked; the content must be served over HTTPS.

This PR changes that so the fonts are retrieved using HTTPS.


lukevella:
Thanks merging.

PR #130, opened 2022-04-23
Add Docker publishing workflow

m3nu:
This adds a simple workflow to build Docker images via Github Actions and push them to Github's container registry.

This will only happen for tagged commits or when run manually from *Actions*. To add a tag:

```
$ git tag -a v0.2
$ git push origin v0.2
```

This could be adjusted to e.g. use SHAs as tags or run more often.

This should automatically use the repo's permissions after merging, so no further setup needed.

Test package here: https://github.com/m3nu/rallly/pkgs/container/rallly

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -1,6 +1,5 @@
 on:
   push:
-  pull_request:
   workflow_dispatch:
     inputs:
       debug_enabled:
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -11,7 +11,8 @@ on:
 jobs:
   build:
     name: Build, push, and deploy
-    #if: ${{ startsWith(github.ref, 'refs/tags/v') }}
+    if: ${{ startsWith(github.ref, 'refs/tags/v') }}
+    #if: github.ref == 'refs/heads/main'
     runs-on: ubuntu-latest
     steps:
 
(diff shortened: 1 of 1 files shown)


.github/workflows/cd.yml
```diff
@@ -0,0 +1,42 @@
...
+      debug_enabled:
+        description: 'Run the build with tmate debugging enabled'
+        required: false
+        default: false
+
+jobs:
+  build:
+    name: Build, push, and deploy
+    if: ${{ startsWith(github.ref, 'refs/tags/v') }}
+    #if: github.ref == 'refs/heads/main'
```
lukevella, line 15: Don't think we need this right?

```suggestion
```

.github/workflows/cd.yml
```diff
@@ -0,0 +1,42 @@
+on:
+  push:
+  pull_request:
```
lukevella, line 3: Why do we want this triggered on pull requests?

lukevella:
Hey @m3nu thanks for setting this up. Really useful!

.github/workflows/cd.yml
m3nu: I left it as example if you want to customize it later.

.github/workflows/cd.yml
m3nu: You probably don't. Sometimes I want to test building the image if changes happen. Let's remove it for now.

.github/workflows/cd.yml
m3nu: E.g. to have a nightly Docker image that's updated when pushing to main.

lukevella (approved):

m3nu:
Great! Waiting for the first tagged release and package to use this official image on PikaPods.

PR #172, opened 2022-05-16
fixed min length of SECRET_PASSWORD.

Dangersohn:
I had the following error in my docker logs:

```bash
rallly-rallly-1 | Error: iron-session: Bad usage. Password must be at least 32 characters long.
rallly-rallly-1 | at file:///usr/src/app/node_modules/iron-session/dist/index.mjs:24:13
rallly-rallly-1 | at Array.forEach (<anonymous>)
rallly-rallly-1 | at getIronSession (file:///usr/src/app/node_modules/iron-session/dist/index.mjs:22:76)
```


lukevella:
Thanks for pointing this out. Not sure where I got 25 from 🤔

PR #182, opened 2022-05-19
docker-compose: pass schema to migrate deploy

osresearch:
This patch passes the explicit path to the prisma schema so that the initial table creation is performed.  Otherwise the command will print:

    No migration found in prisma/migrations

And attempting to create a poll produces the postgres error:

    ERROR:  relation "public.users" does not exist at character 35
    STATEMENT:  SELECT "public"."users"."id" FROM "public"."users" WHERE "public"."users"."email" = $1 OFFSET $2

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -15,7 +15,7 @@ services:
         - DATABASE_URL=postgres://postgres:postgres@rallly_db:5432/db?pgbouncer=true
         - NEXT_PUBLIC_BASE_URL=${NEXT_PUBLIC_BASE_URL}
     restart: always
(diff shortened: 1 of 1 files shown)


lukevella:
Hi, thanks for the PR. I've been looking in to this. The issue is that we're copying the `prisma.schema` into the root of the working directory and then again inside the prisma folder. The first `prisma.schema` is used to get passed installing the dependencies but it makes `prisma migrate deploy` think that the `migrations` folder should also be in the root directory. I think this can be better handled with a [mutli-stage dockerfile](https://docs.docker.com/develop/develop-images/multistage-build/) but this is a good short-term solution. Thanks for your help.

PR #269, opened 2022-07-31
Shrink docker image with multi-stage build

mattprecious:
This reduces the final image from ~2.1gb to ~900mb. I honestly don't really understand node and yarn enough to know what was consuming all that extra space (I assume some sort of build dependency), but the final size of `/usr/src/app` was/is 820mb so the new image size of 900mb checks out.


lukevella:
Thanks @mattprecious, looks good to me.

PR #328, opened 2022-10-11
Remove DATABASE_URL from Dockerfile

### LINKED ISSUE #170: docker builds should not reference a database
https://github.com/lukevella/rallly/blob/357d25305f8162106f363ea2cf81e5f2841b4e53/Dockerfile#L6

Hi, firstly thanks for your work on rallly. However it is not advisable to tie docker builds to a database.
Docker builds should ideally be as idempotent as possible, and database configuration can happen via commands later or entrypoint scripts.

LarsHaalck:
Closes #170

This PR removes the `DATABASE_URL` from the Dockerfile and moves the CMD in the last line to a startup script.
Additionally, the docker compose performs a health check on the database container to wait until is fully functional to avoid refused connections that could sometims occur on the first run when the database was created.

I'm no Prisma expert but this seems to work without the need to override the `datasources` field in the `PrismaClient`. Prisma only complains when the env variable is not set upon startup.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,4 +1,4 @@
-version: "3.3"
+version: "3.9"
 services:
   rallly_db:
     image: postgres:14.2
@@ -7,15 +7,20 @@ services:
       - db-data:/var/lib/postgresql/data
     environment:
       - POSTGRES_PASSWORD=postgres
+      - POSTGRES_DB=db
+    healthcheck:
+      test: ["CMD-SHELL", "pg_isready -U postgres"]
+      interval: 5s
+      timeout: 5s
+      retries: 5
 
   rallly:
     build:
       context: .
-      args:
-        - DATABASE_URL=postgres://postgres:postgres@rallly_db:5432/db?pgbouncer=true
     restart: always
     depends_on:
-      - rallly_db
+      rallly_db:
+        condition: service_healthy
     ports:
       - 3000:3000
     environment:
--- a/docker_start.sh
+++ b/docker_start.sh
@@ -1,0 +1,3 @@
+#!/bin/sh
+yarn prisma migrate deploy --schema prisma/schema.prisma
+yarn start
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,16 +12,14 @@ COPY . .
 
 RUN yarn build
 
-FROM node:alpine	
+FROM node:alpine
 
 ENV PORT 3000
 EXPOSE 3000
 
-ARG DATABASE_URL
-ENV DATABASE_URL $DATABASE_URL
-
 WORKDIR /usr/src/app
 
 COPY --from=build /app .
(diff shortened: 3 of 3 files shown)


docker-compose.yml
```diff
@@ -7,19 +7,21 @@ services:
...
     restart: always
     depends_on:
-      - rallly_db
+      rallly_db:
+        condition: service_healthy
     ports:
       - 3000:3000
-    environment:
-      - DATABASE_URL=postgres://postgres:postgres@rallly_db:5432/db
     env_file:
```
lukevella, line 25: Thanks @LarsHaalck. I think these two lines should stay to avoid having to manually add this to the `.env` file.

LarsHaalck: Okay no problem.

I think the default db, postgres creates inside the container is the same as the `POSTGRES_USER` variable which defaults to `postgres`. So we either have to change the `DATABASE_URL` or add `POSTGRES_DB=db` to the db container. What do you think?

lukevella: I don't think setting `POSTGRES_DB` is enough. `DATABASE_URL` is required to generate the prisma client. It won't work with any other variables.

See here: https://github.com/lukevella/rallly/blob/main/prisma/schema.prisma#L3

lukevella: Ah I see. I wonder how this never caused an issue before because `POSTGRES_DB` was never set and the connection string was always `postgres://postgres:postgres@rallly_db:5432/postgres`. 

Anyway, setting `POSTRES_DB=db` for rallly_db makes sense, but how would the prisma client get the database connection string without setting `DATABASE_URL` as well?

LarsHaalck: I would leave the `DATABASE_URL` and also set `POSTRES_DB=db`.

Interestingly, it also works for me but I get the following error messages when not setting `POSTRES_DB`:
```
rallly-rallly_db-1  | 2022-10-11 09:09:48.915 UTC [75] FATAL:  database "db" does not exist
rallly-rallly_db-1  | 2022-10-11 09:09:48.945 UTC [76] FATAL:  database "db" does not exist
rallly-rallly-1     |
rallly-rallly-1     | PostgreSQL database db created at rallly_db:
```

So it seems, the db is just created when it doesn't exist. I think it would be cleaner to set both environment variables. Let me put the changes in another commit and then you can check again.

lukevella (approved):

lukevella:
LGTM, thanks for your help @LarsHaalck.

docker_start.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
+yarn prisma migrate deploy --schema prisma/schema.prisma
+yarn start
```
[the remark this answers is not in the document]
lukevella: Thanks @haslersn. I've created #369 to keep track of this. Feel free to make a PR to address this.

PR #349, opened 2022-11-04
CI/CD: Automatically publish Docker image

samsapti:
Now that #170 is fixed, we can add a CI/CD workflow to automatically push a Docker image to Dockerhub. Only thing you need to do is to add secrets for your Dockerhub credentials.

### DIFF
--- a/.github/workflows/docker-image.yml
+++ b/.github/workflows/docker-image.yml
@@ -17,7 +17,7 @@ jobs:
           password: ${{ secrets.DOCKER_PASSWORD }}
       
       - name: Check out repository code
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
  
       - name: Build and publish image
         uses: docker/build-push-action@v3
--- a/.github/workflows/docker-image.yml
+++ b/.github/workflows/docker-image.yml
@@ -1,5 +1,4 @@
-name: Build and push Docker image
-
+name: Build and publish Docker image
 on:
   push:
     branches:
@@ -7,11 +6,11 @@ on:
   workflow_dispatch:
 
 jobs:
-  build-push:
-    name: Build and push image
+  publish:
+    name: Build and publish Docker image
     runs-on: ubuntu-latest
     steps: 
-      - name: Login to DockerHub
+      - name: Log in to Docker Hub
         uses: docker/login-action@v2
         with:
           username: ${{ secrets.DOCKER_USERNAME }}
@@ -20,7 +19,7 @@ jobs:
       - name: Check out repository code
         uses: actions/checkout@v2
  
-      - name: Build and push image
+      - name: Build and publish image
         uses: docker/build-push-action@v3
         with:
           push: true
--- a/.github/workflows/docker-image.yml
+++ b/.github/workflows/docker-image.yml
@@ -1,0 +1,29 @@
+name: Build and push Docker image
+
+on:
+  push:
+    branches:
+      - main
+  workflow_dispatch:
+
+jobs:
+  build-push:
+    name: Build and push image
+    runs-on: ubuntu-latest
+    steps: 
+      - name: Login to DockerHub
+        uses: docker/login-action@v2
+        with:
+          username: ${{ secrets.DOCKER_USERNAME }}
+          password: ${{ secrets.DOCKER_PASSWORD }}
+      
+      - name: Check out repository code
+        uses: actions/checkout@v2
+ 
+      - name: Build and push image
+        uses: docker/build-push-action@v3
+        with:
+          push: true
+          tags: |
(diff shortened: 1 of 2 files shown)


.github/workflows/docker-image.yml
```diff
@@ -0,0 +1,29 @@
+name: Build and push Docker image
```
justinmayer, line 1: Suggested re-wording: `Build and publish Docker image`

.github/workflows/docker-image.yml
```diff
@@ -0,0 +1,29 @@
...
+
+on:
+  push:
+    branches:
+      - main
+  workflow_dispatch:
```
justinmayer, line 7: Nice work on including `workflow_dispatch`, so this Docker build+publish action can be [performed manually within GitHub's UI](https://github.blog/changelog/2020-07-06-github-actions-manual-triggers-with-workflow_dispatch/) 👍

.github/workflows/docker-image.yml
```diff
@@ -0,0 +1,29 @@
...
+
+jobs:
+  build-push:
```
justinmayer, line 10: Suggested name for this job: `publish`

.github/workflows/docker-image.yml
```diff
@@ -0,0 +1,29 @@
...
+    name: Build and push image
```
justinmayer, line 11: Suggested re-wording: `Build and publish Docker image`

.github/workflows/docker-image.yml
```diff
@@ -0,0 +1,29 @@
...
+    runs-on: ubuntu-latest
+    steps: 
+      - name: Login to DockerHub
```
justinmayer, line 14: Suggested re-wording: `Log in to Docker Hub`

.github/workflows/docker-image.yml
```diff
@@ -0,0 +1,29 @@
...
+        uses: docker/login-action@v2
+        with:
+          username: ${{ secrets.DOCKER_USERNAME }}
+          password: ${{ secrets.DOCKER_PASSWORD }}
+      
+      - name: Check out repository code
+        uses: actions/checkout@v2
+ 
+      - name: Build and push image
```
justinmayer, line 23: Suggested re-wording: `Build and publish image`

justinmayer (approved):
Great work, Sam! I made a few *very* minor wording suggestions, just in case you find them useful. Excited to have this merged! ✨

samsapti:
@justinmayer I like your suggestions :+1:

samsapti:
I've also added a cron schedule to rebuild the image once a week. This is to keep the image (dependencies, etc.) up-to-date.

lukevella:
Thanks for doing this! I’ll try and test this out asap but might have to wait till Monday.

> I've also added a cron schedule to rebuild the image once a week. This is to keep the image (dependencies, etc.) up to date.

I don’t think a cron schedule is necessary. The dependencies are locked with a lock file there shouldn’t be any difference in the dependencies if the commit is the same.

samsapti:
> I don’t think a cron schedule is necessary. The dependencies are locked with a lock file there shouldn’t be any difference in the dependencies if the commit is the same.

@lukevella what I meant was for the Docker image itself to be kept up to date, meaning the base image (`node:alpine`) + Node.js, Yarn, etc. I can remove it if you want me to.

lukevella:
> @lukevella what I meant was for the Docker image itself to be kept up to date, meaning the base image (`node:alpine`) + Node.js, Yarn, etc. I can remove it if you want me to.

Yes, I think it should be removed. Producing different results for the same commit would make debugging harder. Ideally, all dependency versions are pinned to avoid this from happening but that can be addressed separately.

samsapti:
@lukevella done (force-pushed last commit but without cron schedule) :+1:

.github/workflows/docker-image.yml
```diff
@@ -0,0 +1,28 @@
...
+    runs-on: ubuntu-latest
+    steps: 
+      - name: Log in to Docker Hub
+        uses: docker/login-action@v2
+        with:
+          username: ${{ secrets.DOCKER_USERNAME }}
+          password: ${{ secrets.DOCKER_PASSWORD }}
+      
+      - name: Check out repository code
+        uses: actions/checkout@v2
```
valberg, line 20: checkout is at v3 now, better use the newest version

valberg (changes requested):

.github/workflows/docker-image.yml
samsapti: Fixed, thanks!

lukevella:
Thanks, everyone for your help. Looks good to me!

PR #379, opened 2023-01-03
Fix the enabling of Postgres extension "citext"

scottwallacesh:


lukevella:
Hi thanks for pointing this out. I think editing migration files like this might not work but not sure how else to solve this right now. Keeping an eye on this.

scottwallacesh:
I needed this modification because I was installing it via Docker into an existing Postgres instance and didn't want to provide admin-level access to the user.

This modification allowed me to create an empty DB, enable the extensions in it, create a non-admin user and then start the app with that user and have it work.

lukevella (approved):

lukevella:
Should be fine to merge but will do it next week so I can keep an eye on things in case something goes wrong.

PR #410, opened 2023-01-19
Set supportEmail when rendering login template

gramakri:
The contact link is rendered as "mailto:undefined" in the magic link emails


lukevella:
Thank you for your help!

PR #425, opened 2023-01-21
Update sentry

lukevella:


lukevella:
Updating to `@sentry/nextjs@7.31.1` was causing the build to fail. Not sure what's wrong but might be worth waiting to see if a future update fixes this issue.

PR #490, opened 2023-02-07
bump trpc

KATT:
seems to work on my machine


lukevella:
Closing this since it's the same result as #489

PR #519, opened 2023-02-22
Improve time slot start times

jonas-hoebenreich:
Set start time of new time slot to end of previous slot

Currently, the time slots all start at the same time. This is cumbersome and creates additional effort when creating polls.
This pull request sets the start time of the new time slot to the end time of the previous time slot. This allows users to create successive time slots with just a few clicks.

see https://github.com/lukevella/rallly/discussions/138


lukevella:
Thanks @jonas-hoebenreich, originally there were some issues in doing it this way when the new start time when past midnight but having tested it again it seems to work fine so happy to merge this. Thanks again for your contribution.

PR #553, opened 2023-03-13
✨ Allow users to log in with magic link

lukevella:


apps/web/src/utils/auth.ts
```diff
@@ -33,6 +33,7 @@ export type RegistrationTokenPayload = {
 export type LoginTokenPayload = {
   userId: string;
   code: string;
+  redirect?: string;
```
lukevella, line 36: Remove this

PR #561, opened 2023-03-14
💦 Temporarily disable arm64 builds

lukevella:
The reason why this is necessary is because our pipeline is failing due to network timeouts. Unfortunately I am not aware of a solution to this problem but if we can find a way to build these images as part of our workflow we should reeanble this feature.


vitkolos:
Are there any plans to enable the arm64 builds again?

lukevella:
See https://github.com/lukevella/rallly/discussions/731

PR #564, opened 2023-03-14
💄 Hide comment dropdown menu when unavailable to user

jonas-hoebenreich:
Currently, each comment has a menu, with a button to delete the comment. If the user is not allowed to delete the comment, the button is disabled (see first screenshot).
With this pull request the menu button is only shown if necessary. This reduces clutter and improves the user experience.

Before:
image

After:
image

### DIFF
--- a/apps/web/src/components/discussion/discussion.tsx
+++ b/apps/web/src/components/discussion/discussion.tsx
@@ -120,7 +120,6 @@ const Discussion: React.FunctionComponent = () => {
                       <DropdownItem
                         icon={Trash}
                         label={t("deleteComment")}
-                        disabled={!canDelete}
                         onClick={() => {
                           deleteComment.mutate({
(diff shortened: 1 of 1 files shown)


apps/web/src/components/discussion/discussion.tsx
```diff
@@ -112,22 +112,24 @@ const Discussion: React.FunctionComponent = () => {
...
-                  </Dropdown>
+                  {canDelete &&
+                    <Dropdown
+                      placement="bottom-start"
+                      trigger={<CompactButton icon={DotsHorizontal} />}
+                    >
+                      <DropdownItem
+                        icon={Trash}
+                        label={t("deleteComment")}
+                        disabled={!canDelete}
```
lukevella, line 123: We can remove the `disabled` prop here since it will always be `false`.
```suggestion
```

PR #600, opened 2023-03-22
Enable zooming and scaling

jonas-hoebenreich:
According to a11y standards, zooming and scaling must not be disabled, because people with low vision rely on screen magnifiers to properly see the contents of a web page.
see https://dequeuniversity.com/rules/axe/4.4/meta-viewport

I see no reason to disable zooming and scaling, so I've modified this accordingly.


lukevella:
Thanks @jonas-hoebenreich, good catch.

PR #601, opened 2023-03-23
✨ Show version number in app

### LINKED ISSUE #588: Display version number somewhere on the app
Generally useful to be able to see what version is running.

## To do:
- [ ] Add a new environment `NEXT_PUBLIC_APP_VERSION`
- [ ] Update `apps/web/Dockerfile`  to pull the version from the root `package.json` and set this environment variable
- [ ] Show the value of this variable somewhere in the app

arcticFox-git:
Resolves #588  - Display version number somewhere on the app

- Added two lines to the dockerfile that pull the version from the root `package.json` file and add it to an `.env` file that next.js can pull from. This took a while to understand, the docker container environmental variables pulls from the root `.env` but next.js seems to like an `.env` file in `app/apps/web` (at least in the way that I was using it).
- Adjusted `footer.tsx` to add the version number (in the form vX.X.X so as to be language agnostic) below the RALLLY logo. 
- If the value for the `NEXT_PUBLIC_APP_VERSION` environment variable is undefined, then the app will look like normal without anything additional showing. This pertains to people running `yarn run dev`, as then next.js appears to pull from the root `.env` instead, and so if they do not set this in their `.env` then it would be undefined.

Overall, I was a little confused with the `.env` scope based on the README. The `sample.env` is located in apps/web/ however the README says to copy it using `cp sample.env` after `cd` into the rallly directory. 

This is my first pull request to a public repo, so open to feedback on how I went about this process and if I have done anything incorrectly.


lukevella:
I've been having a look at this and I can see it's a bit more complicated than I originally anticipated so don't worry if you need to drop this. I'll be looking into it soon.

arcticFox-git:
Hi @lukevella, yeah I might leave it to you from here, but I did learn a lot from working through this so very grateful for that! If it helps, below is the gist of what worked and what didn't. 

If NEXT_PUBLIC_APP_VERSION is set as an environment variable in the root directory, it will successfully display in the app through yarn dev, yarn build etc, but not through docker build. If it is set in the apps/web directory, it will work through docker build but not through yarn. 

Feel free to delete this PR at your convenience!

lukevella (approved):

lukevella:
I think I've found a solution for this though it's not quite what I originally had in mind. Rather than trying to set the value of `NEXT_PUBLIC_APP_VERSION` from within the Dockerfile I am setting this value through a build argument.

Thanks for your help with this @arcticFox-git.

PR #605, opened 2023-03-23
✨ Prefill participant form if user is logged in

lukevella:


lukevella:
Not sure if this is still a good idea to be honest. Gna table it.

PR #607, opened 2023-03-23
♿️ Improve contrast for secondary text

jonas-hoebenreich:
turn text-slate-400 into text-slate-500
While text-slate-400 certainly looks elegant on a light background, it is difficult to read for people with limited vision. I have therefore replaced all places with text-slate-400 with text-slate-500. The visual change is minimal, but the contrast now meets a11y standards (see https://color.a11y.com/)

### DIFF
--- a/apps/web/src/components/auth/login-form.tsx
+++ b/apps/web/src/components/auth/login-form.tsx
@@ -94,7 +94,7 @@ const VerifyCode: React.FunctionComponent<{
               {formState.errors.code.message}
             </p>
           ) : null}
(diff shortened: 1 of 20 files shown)


lukevella:
Thanks for pointing this out. Having looked around a bit, I'm not missing the light colored text as much as I thought I would. I'll give this a more thorough look once I have the chance 👍

lukevella:
Hey @jonas-hoebenreich, I'm going to close this and continue it in #613 which increases the contrast for pretty much all the colors and not just the light text. Thanks for raising this, it's something I've been meaning to address.

PR #608, opened 2023-03-23
Dark mode

jonas-hoebenreich:
Introduce Dark Mode

The dark mode probably still needs a bit of fine-tuning in some corners :)


lukevella:
Hey @jonas-hoebenreich, thanks for sharing this, it's really interesting to see. I probably won't merging this though because it would take quite a bit effort to get this ready for production and it also has implications on the maintainability of the app going forward. My hands are full even with just one theme to think about. Dark mode will come but it's not time yet.

PR #623, opened 2023-03-29
Add discussion option to poll

TheZoker:
## Description

This adds a new option, that allows a user to enable/disable the comments/discussion section below the poll:

image

Maybe using a switch component for the toggle would be better, but I was not sure how to do that.

## Checklist

### DIFF
--- a/apps/web/public/locales/de/app.json
+++ b/apps/web/public/locales/de/app.json
@@ -45,6 +45,7 @@
   "email": "E-Mail",
   "emailNotAllowed": "Diese E-Mail ist nicht erlaubt.",
   "emailPlaceholder": "max.mustermann@mail.de",
+  "enableDiscussion": "Kommentare erlauben",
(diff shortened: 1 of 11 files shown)


lukevella:
re: comments

My plan is to change how this feature works altogether. Rather than having comments for discussion, participants will have the option to leave a comment or note to the poll creator when submitting a response. The comments will only be visible to the poll creator. No need to have an option to disable this.

re: limiting number of participants per option

I've thought long and hard about this and the conclusion I've reached is that the only purpose for this feature is to use meeting polls as booking pages. That's not something I want to encourage as it just makes things more confusing. Maybe I will offer booking pages as a separate solution to the meeting polls but that's far down the road.

Regarding ETAs, I don't want to make any promises. I'm quite burnt out at the moment and need to slow down to recover.

TheZoker:
Hi @lukevella 

Thanks for the heads up!
Sounds good. 

> I'm quite burnt out at the moment and need to slow down to recover.

Take your time and foremost take care of yourself! Hope you get better soon

PR #628, opened 2023-03-29
Add default language option

TheZoker:
## Description

This PR should allow to set a different default language. 
For some reason I was not able to test this, since the env was always undefined for me. But I could not find any other place, where I need to add it.

## Checklist


lukevella:
This doesn't do what you think it's doing. Rallly uses Next.js locale detection and similar logic in [middleware](https://github.com/lukevella/rallly/blob/main/apps/web/src/middleware.ts) to decide what language to show the user. This is based on the `accept-language` header which is configured in the users browser. So if a user prefers german language, by default they will already be served the german language.

`defaultLocale` needs to be a locale that is not missing any translation strings because this is the fallback language when certain translation keys are missing and when no matching locales could be found from the `accept-language` header.

Additionally, I don't believe this will work for users running the prebuilt docker image because `defaultLocale` is only used at build time.

PR #653, opened 2023-04-05
add new email providers sendgrid, mailgun, sendinblue, mailjet

cedricwalter:
## Description

add ses env config
improved sample.env
add new providers sendgrid, mailgun, sendinblue, mailjet

## Checklist

### DIFF
--- a/apps/web/declarations/environment.d.ts
+++ b/apps/web/declarations/environment.d.ts
@@ -92,6 +92,30 @@ declare global {
        * The app version just for reference
        */
       NEXT_PUBLIC_APP_VERSION?: string;
+      /**
+      * Sendgrid API Key
+      */
+      SENDGRID_API_KEY?: string;
+      /**
+      * Send In Blue API Key
+      */
+      SENDINBLUE_API_KEY?: string;
+      /**
+      * Mailjet API Key
+      */
+      MAILJET_API_KEY?: string;
+      /**
+      * Mailjet API Key
+      */
+      MAILJET_API_SECRET?: string;
+      /**
+      * Mailgun API Key
+      */
+      MAILGUN_API_KEY?: string;
+      /**
+      * Mailgun API Key
+      */
+      MAILGUN_DOMAIN?: string;
     }
   }
 }
--- a/packages/emails/src/send-email.tsx
+++ b/packages/emails/src/send-email.tsx
@@ -73,51 +73,51 @@ const getTransport = () => {
       }
       break;
     case "mailgun":
-          {
-           const mailgunTransport = require('nodemailer-mailgun-transport');
+      {
+       const mailgunTransport = require('nodemailer-mailgun-transport');
 
-           // This is your API key that you retrieve from www.mailgun.com/cp (free up to 10K monthly emails)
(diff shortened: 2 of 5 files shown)


packages/emails/src/send-email.tsx
```diff
@@ -49,6 +49,75 @@ const getTransport = () => {
         });
       }
       break;
+    case "sendgrid":
+      {
+        const nodemailerTransport = require('nodemailer-sendgrid-transport');
```
lukevella, line 54: Can you use import instead of require since this is an es module?

sample.env
```diff
@@ -1,9 +1,60 @@
...
 SMTP_PORT=587
 SMTP_SECURE=false
 SMTP_USER=your-smtp-user
 SMTP_PWD=your-smtp-password
+
+## Custom email provider (higher priority, uncomment one to activate)
+# Amazon ses
+#EMAIL_PROVIDER=ses
+#AWS_REGION=
+#API_KEY=
```
lukevella, line 40: Some call it api key and others call it access key. There's an opportunity here to be less ambiguous by defining separate env vars for each of these providers.

lukevella (commented):
`environment.d.t.s` will need to be updated as well with any new environemnt variables and any changes we make.

apps/web/declarations/environment.d.ts
lukevella, line 1: We should update `EMAIL_PROVIDER` with the new values here as well otherwise it should throw ts errors in the emails package.

It's possible these types aren't currently avialble in that package right now but I'll work on fixing that.

packages/emails/src/send-email.tsx
```diff
@@ -5,6 +5,12 @@ import { createTransport, Transporter } from "nodemailer";
 import type Mail from "nodemailer/lib/mailer";
 import React from "react";
 
+import { sendgridTransport } from "nodemailer-sendgrid-transport";
+import { mandrillTransport } from "nodemailer-mandrill-transport";
+import { mailgunTransport } from "nodemailer-mailgun-transport";
+import { sendinblueTransport } from "nodemailer-sendinblue-transport";
+import { mailjetTransport } from "nodemailer-mailjet-transport";
```
lukevella, line 12: I think importing these on demand as you had done with require was a good idea to keep the initial bundle size smaller.

Let's try using import inline inside the switch statement instead.

```
const { sendgridTransport } = await import("nodemailer-sendgrid-transport");
```

lukevella:
Hi @cedricwalter, thanks for working on this. Could you let me know if you've tested against any of these providers?

packages/emails/src/send-email.tsx
```diff
@@ -49,6 +55,66 @@ const getTransport = () => {
         });
       }
       break;
+    case "sendgrid":
+      {
+        const options = {
```
lukevella, line 60: Make sure you have prettier running to auto-fix the code style to match the rest of the codebase.

lukevella:
Hi @cedricwalter, I'm going to close this since there hasn't been any activity on it. If you wish to continue working on it feel free to reopen.

PR #663, opened 2023-04-15
Add docker secrets support

Flash1232:
## Description

This PR aims to add support for docker secrets for sensitive config values such as the app secret and the SMTP credentials.

## Checklist

- [ ] I have performed a self-review of my code
- [ ] I have commented my code, particularly in hard-to-understand areas


lukevella:
Hey @Flash1232, thanks for sharing this with me. My take is that adding the feature might complicate things unnecessarily, and since there hasn't been much demand for it, I think we should hold off on it for now. But if you can give me a good reason to reconsider, I'm open to it.

Flash1232:
Hi @lukevella,
thanks for your insight! The main reason for this is to avoid the risk of having unencrypted secrets on the host's disk but rather let Docker (Swarm) take care of passing it directly to the container.
Of course it is up to you whether to accept this PR or not and I understand it might add some complexity for users when first reading the configuration docs.

Many docker images offer these additional `_FILE` variables, one example: https://github.com/anonaddy/docker#app (as described in the note).

Feel free to close the PRs if you feel this is unwanted or otherwise does not suit your needs as of now. Thank you!

lukevella:
Thanks @Flash1232, that's an excellent example. I like how they documented it in a note so that it doesn't overload the rest of the config. I'd probably choose to do the same.

I'd like to take some time to consider whether this is worth doing right now since it's not something that has been raised before. I'll close this PR but thanks for raising it. If you wish to open a [new discussion](https://github.com/lukevella/rallly/discussions/new?category=ideas), that would be a good starting point to start gauging interest for this feature.

Issue #761, opened 2023-07-15
Error starting up locally


kleenkanteen:
**Describe the bug**
Initially my typescript version was wrong, a bit higher than the workspace version. But I changed it to the workspace version and still get the same errors.

I get these 2 errors when running `yarn dev`. Here is the terminal output of the first error:

![image](https://github.com/lukevella/rallly/assets/67878058/66f2c6df-f441-4d97-9b1c-94f786fa33e4)

Here is the second error:

![image](https://github.com/lukevella/rallly/assets/67878058/17f54b05-836d-49ff-b45f-0856c3e759f5)

I created a "static" folder under rallly\packages\emails\src\templates\ and reran `yarn dev` but same issue.

Here is the full terminal output with both errors. The first one starts 1/5th of the way down, with the command being "email dev --port 3333 --dir ./src/templates". The second error starts 3/5th of the way down with "TypeError [ERR_INVALID_URL]: Invalid URL":
![image](https://github.com/lukevella/rallly/assets/67878058/559d615f-4528-4384-9690-ef2462162d52)

**To Reproduce**
1. Follow the development guide on the readme.
2. At the step `yarn dev`, I get this error.

**Expected behavior**
@rallly/emails:dev server and @rallly/landing:dev server to start up normally.

**Screenshots**
For the error saying `@rallly/landing:dev: TypeError [ERR_INVALID_URL]: Invalid URL`, here is the line of code it seems to break on, line 49:

![image](https://github.com/lukevella/rallly/assets/67878058/47ed1517-a4df-45ef-8a4d-b7ee372885b3)

rallly\apps\landing\next.config.js:49:22

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser: none so far
 - Version: 2H22, latest version

**Additional context**

kleenkanteen:
For the email issue, this is a related issue to my problem: https://github.com/resendlabs/react-email/issues/613

lukevella:
Any idea if updating react-email fixes this?

kleenkanteen:
react-email was on the latest version, 1.9.4.

I tried setting up rallly on WSL, but I couldn't get WSL to connect to postgres on the host windows.

I then installed Kubuntu in a VM. I followed the developement guide. Then at yarn dev, I get pretty the same 2 errors I got on windows:

![image](https://github.com/lukevella/rallly/assets/67878058/8ac61d0d-3a58-46b6-b72e-c35a85a8549e)

[...]

PR #774, opened 2023-07-20
Fix react email version

kleenkanteen:
## Description

Fix the react-email issue I had in #761 by fixing the version to a older one.

## Checklist

- [X] I have performed a self-review of my code
- [X] My code follows the code style of this project
- [X] I have commented my code, particularly in hard-to-understand areas

### DIFF
--- a/packages/emails/package.json
+++ b/packages/emails/package.json
@@ -16,7 +16,7 @@
     "@react-email/render": "0.0.6",
     "clsx": "^1.2.1",
     "nodemailer": "^6.9.1",
-    "react-email": "^1.9.1"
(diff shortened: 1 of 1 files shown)


lukevella:
Not sure if downgrading makes sense here. The error complaining about a missing static folder is annoying but doesn't actually break anything AFAIK. We could just add the static folder to silence the error. Not ideal but much prefer that then not being able to upgrade this package.

kleenkanteen:
I tried adding a static folder but even then the error occured.

If it doesn't break anything no problem then. I'll close this PR.

Issue #697, opened 2023-05-31
Authentication with oauth


kolaente:
It would be very cool if rallly provided a way to authenticate using a third party auth provider using the oauth / openid protocol.

icsy7867:
Seconded - OIDC, SAML, Or HTTP Header authentication would be great.

mpldr:
Thirded and adding that OIDC is probably relatively easy to implement using an existing library for it, if one does know JS and how OAuth works. I do not know JS, unfortunately.

lukevella:
First step is to start using next-auth. However it doesn't have any providers that support anonymous/guest sessions so will likely need to roll our own and we'll need to make ensure compatibility with the existing sessions so that existing guests don't get locked out of their polls or responses.

mpldr:
> However it doesn't have any providers that support anonymous/guest
> sessions so will likely need to roll our own and we'll need to ensure
> compatibility with the existing sessions

Again, I may be showing my ignorance on how the JS-crowd does things, so
bear with me :D

Right now users have a token of some kind that identifies them as the
owner/administrator of a poll. If someone with such a token signs in
through an SSO provider, the identifier and the account it should be
linked to are known, so would it not be easy™ to migrate polls to the
newly created account?

Go-ey pseudocode:

```
if cookies.IsSet("account-cookie") && cookies.IsSet("guest-session") {
	database.ChangeOwner(cookies.Get("guest-session"), cookies.Get("account-cookie"))
	cookies.Delete("guest-session")
}

var activeIdentity uuid.UUID

if cookies.IsSet("account-cookie") {
	activeIdentity = cookies.Get("account-cookie")
}

if cookies.IsSet("guest-session") {
	activeIdentity = cookies.Get("guest-session")
}

// do permission stuff and the like
```

Again, as I am not a JS dev and rather unfamiliar with the codebase,
I can only act as somewhat of a rubber-duck here. Feel free to use me
for that purpose though, if you so desire :D

-- 
Moritz Poldrack
https://moritz.sh

> Breaking seal constitutes acceptance of agreement.

lukevella:
I wasn't referring to migrating polls from guests to users. Even more basic then that, changing how authentication works could invalidated all sessions. That means:

* Everyone will be logged out (not a big deal, just log in again)
* All guest session will be renewed

[...]

PR #939, opened 2023-11-18
Support openid connect for login

armanddidierjean:
close: #697

## Description

Hello,

As next-auth is now used for authentication, it is possible to support oauth and oidc login without requiring an external library. In this pull request, I added support for oauth and oidc.

## How does it work?
There are four configuration variables that can be set to enable oidc/oauth login:
```
OIDC_NAME
OIDC_DISCOVERY_URL
OIDC_CLIENT_ID
OIDC_CLIENT_SECRET
```

When they are set, a new Oauth provider is added in `apps/web/src/utils/auth.ts`.

> I think Github display strangelly the changes I did to `getAuthOptions` at the begining of `apps/web/src/utils/auth.ts` file.
> I created an array `providers` at the top of the file, to make it easy to conditionnally add a new provider if it is configured in the dotenv.
> Then I passed the new array in place of the old one in `getAuthOptions`

I added some logic in `apps/web/src/utils/auth.ts` `sign-in` callback that create a new user or link an existing one when someone log in using oidc/oauth.

The login form now displays a list of configured oauth providers, even though this PR only allows for one to be configured.
```
        {providers ? <hr className="border-t border-grey-500 my-2" /> : null}

        {providers &&
          Object.values(providers).map((provider) =>
            provider.type === "oauth" ? (
              <Button
                size="lg"
                asChild
                onClick={() => signIn(provider.id)}
                key={provider.name}
              >
                <Link href="">Se connecter avec {provider.name}</Link>
              </Button>
            ) : null,
          )}
```
Capture d’écran 2023-11-18 à 09 32 25

## Checklist

If you think these change may be interesting to add to Rallly, I will be happy to improve the pull request based on received feedback and suggestions.


troplin:
This would be fantastic! This will create completely separate internal accounts from the previous email based ones, right?

lukevella:
Hey @armanddidierjean, thanks for this. I've pushed some changes to your branch to fix some linting issues and prettier styling. It would help to have these tools running as it would fail in the CI process otherwise. On first look it looks good. I'll do a proper review soon and hopefully we'll get this merged and released in the next couple of days.

armanddidierjean:
Oh, sorry, I missed the linting, I will make sure to run it next time.

> This would be fantastic! This will create completely separate internal accounts from the previous email based ones, right?
@troplin In this PR it links (using email) existing Rallly accounts with the one given by the provider. 
As emails should be unique in the database, I'm not sure how we could create completely separated accounts.

troplin:
> @troplin In this PR it links (using email) existing Rallly accounts with the one given by the provider. As emails should be unique in the database, I'm not sure how we could create completely separated accounts.

As much as I would like to have OIDC support, I don't think this is a good idea. With OIDC, users can change their email address and this should not result in a new account.

See also https://next-auth.js.org/faq#security

armanddidierjean:
> As much as I would like to have OIDC support, I don't think this is a good idea. With OIDC, users can change their email address and this should not result in a new account.

This won't result in a new account. The flow is the following:

 - if the provider `account` is already linked with a Rallly `user`: log in with this account
 - else:
     - if there is already a Rallly `user` with the same email, we link the `user` with the provider `account`
     - else we create a new `user` and link it with the provider `account`

However, I agree there may be some debate for the first connection. If there is already a Rallly user with the same address email, do we want them to be linked?
I think it would not be easy to create a new distinct account as the email is already used. Moreover, in the case of Rallly which don't use password based login but unique code send by email, I'm not really shocked by the idea of linking the accounts.

lukevella:
I think the security concern is only due to the fact that an identity provider may authenticate a user with an email that hasn't been verified. In theory you could set up your own IDP that allows you to authenticate a user with an email that you do not own and "hijack" an existing account, however I don't think this is a concern in this case since self-hosters would be using their own IDP which is trusted (presumably).

apps/docs/self-hosting/configuration-options.mdx
```diff
@@ -68,3 +68,56 @@ These variables need to be configured to let Rallly send out transactional email
 <ParamField path="SMTP_TLS_ENABLED" default={"false"}>
   Enable TLS for your SMTP connection
 </ParamField>
+
+### Openid Connect Configuration
+
+These variables needs to be configured to enable oauth 2.0 or openid connect (oidc) log in.
+Rallly use [NextAuth.js](https://next-auth.js.org/) for authentication and most options are derived from the framework.
```
lukevella, line 75: Next-auth is an implementation detail that does not need to be exposed to self-hosters. Can you remove these lines? These docs should be self-sufficient or at most point to docs that are referring to open standards like oauth 2.0 or openid.

apps/web/src/utils/auth.ts
```diff
@@ -109,26 +109,37 @@ const providers: Provider[] = [
...
+      url: process.env.OIDC_USER_INFO_URL,
+    };
+  }
   providers.push({
     id: process.env.OIDC_NAME,
     name: process.env.OIDC_NAME,
     type: "oauth",
     wellKnown: process.env.OIDC_DISCOVERY_URL,
-    authorization: { params: { scope: "openid" } },
+    authorization: { params: { scope: process.env.OIDC_SCOPES ?? "openid" } },
```
lukevella, line 132: The scope should be set by the app. We need `openid email profile` as otherwise this would not work.

apps/web/declarations/environment.d.ts
```diff
@@ -80,6 +80,25 @@ declare global {
...
        */
       OIDC_CLIENT_SECRET?: string;
+      /**
+       * Scopes that should be used when configuring the oidc provider
+       */
+      OIDC_SCOPES?: string;
+      /**
+       * If Rallly should expect the oidc provider to return an ID token
+       */
+      OIDC_ID_TOKEN_EXPECTED?: boolean;
```
lukevella, line 90: Currently environment variables are not parsed so they will all be strings.

apps/web/src/utils/auth.ts
```diff
@@ -109,26 +109,37 @@ const providers: Provider[] = [
...
     clientId: process.env.OIDC_CLIENT_ID,
     clientSecret: process.env.OIDC_CLIENT_SECRET,
-    idToken: true,
+    idToken: process.env.OIDC_ID_TOKEN_EXPECTED ?? true,
```
lukevella, line 135: Since `OIDC_ID_TOKEN_EXPECTED` is going to be a string this should be

```
process.env.OIDC_ID_TOKEN_EXPECTED  !== "false"
```

This will default the value to `true` unless explicitly disabled.

apps/web/src/utils/auth.ts
```diff
@@ -109,26 +109,37 @@ const providers: Provider[] = [
 
 // If we have an OAuth provider configured, we add it to the list of providers
 if (process.env.OIDC_NAME) {
+  let userinfo = {};
+  if (process.env.OIDC_FORCE_USER_INFO) {
+    userinfo = {
+      ...userinfo,
+      async request(context: any) {
```
lukevella, line 116: Next-auth docs suggest not using `userinfo` in this this way. 

https://next-auth.js.org/configuration/providers/oauth#userinfo-option

I think it's better to keep it simple and assume that the identity provider is openid connect compliant and will return the data we need in the id token. This means `idToken` should always be `true` and `userinfo` is not necessary.

It is much easier to add configuration down the line then to remove configuration.

apps/web/declarations/environment.d.ts
armanddidierjean: I didn't know, thanks for the information. I changed the two variables that were previously boolean.

apps/web/src/utils/auth.ts
armanddidierjean: This is a good idea, even if some provider may not use `email profile` most should. This is a good default value, thanks.

apps/web/src/utils/auth.ts
armanddidierjean: To be honnest, I am a bit perplexed with next-auth configuration for `userinfo`

`idToken` may not be true using an OAuth 2.0 instead of an openid connect provider. If `idToken` is false, `userinfo` will be called by NextAuth. This make sense.

---

> "The Claims requested by the profile, email, address, and phone scope values are returned from the UserInfo Endpoint, as described in [Section 5.3.2](https://openid.net/specs/openid-connect-core-1_0.html#UserInfoResponse), when a response_type value is used that results in an Access Token being issued. However, when no Access Token is issued (which is the case for the response_type value id_token), the resulting Claims are returned in the ID Token."

According to [openid specifications](https://openid.net/specs/openid-connect-core-1_0.html#ScopeClaims), if I understood them correctly, requested claims should be provided by `userinfo` endpoint if the `response_type` return an access_token. This is the case with `response_type=code` used by next-auth and most oidc clients.

In practice, I think there are clients using one or the other method independently of the specifications. I believe there would be benefits to supports getting claims from both `userinfo` and `id_token`.

armanddidierjean: I could not make `OIDC_USER_INFO_URL` works with `OIDC_FORCE_USER_INFO` but I don't think this is really an issue as `OIDC_USER_INFO_URL` should be used for oauth 2.0 providers (that don't have a discovery_url) while `OIDC_FORCE_USER_INFO` are only useful for some oidc providers.

lukevella:
Hey @armanddidierjean, I pushed some changes to your branch. Here's a summary:

* Mapped names in accounts table to follow naming conventions.
* Reduce the amount of config. Only bare minimum and what is expected for OIDC compliant providers. Supporting all OAuth provider is not something I am interested in.
* Updated UI for login and registration to support additional providers in the future
* Use `allowDangerousAccountLinking` for linking accounts. No need to do this manually.
* Added `OIDC_ENABLED` environment variable to explicitly turn on OIDC login instead of required `OIDC_NAME` which is now optional.
* Added documentation for required scopes and callback url.

I'll wait for your feedback to see if there's any further changes but otherwise this seems like it's pretty much ready to merge to me.

lukevella:
This looks pretty stable to me so going to go ahead and merge it. Support for specific OAuth providers (ex. Google, Azure Active Directory B2C etc…) will follow but the current scope is to allow self-hosters to use their own OIDC compliant providers.

Thanks for your help with this @armanddidierjean.

PR #1126, opened 2024-05-28
Remove unused dependency react-email

jirutka:
## Description

This package is not imported anywhere and has a lot of transitional runtime dependencies. Some of these are clearly build/development dependencies (e.g. swc, webpack) and they are even propagated into NextJS standalone builds, making it much bigger than it needs to be.

Note that there're also @react-email/components and @react-email/render - unlike "react-email", these dependencies are used.

## Checklist

## Summary by CodeRabbit

- **Chores**
  - Removed unused dependency `"react-email": "^2.0.0"` from the email package.

### DIFF
--- a/packages/emails/package.json
+++ b/packages/emails/package.json
@@ -15,8 +15,7 @@
     "@react-email/components": "^0.0.14",
     "@react-email/render": "^0.0.12",
     "nodemailer": "^6.9.9",
-    "preview-email": "^3.0.19",
-    "react-email": "^2.0.0"
+    "preview-email": "^3.0.19"
   },
   "devDependencies": {
     "@rallly/tailwind-config": "*",
--- a/yarn.lock
+++ b/yarn.lock
@@ -1516,13 +1516,6 @@
   dependencies:
     regenerator-runtime "^0.14.0"
 
-"@babel/runtime@^7.23.5":
-  version "7.23.9"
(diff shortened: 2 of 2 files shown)


lukevella:
The `react-email` package is used to be able to [launch a dev server](https://github.com/lukevella/rallly/blob/main/packages/emails/package.json#L6) (amongst other things) that allow us to preview our email templates during development.

jirutka:
Isn't this provided by `preview-email`? Anyway, then it should be in `devDependencies`, not `dependencies`.

lukevella:
> Isn't this provided by preview-email? 

`preview-email` serves a different purpose. It opens the an email in the browser during development. Useful for things like getting a verification code but it's not for building email templates.

> Anyway, then it should be in devDependencies, not dependencies

Maybe, but the [react-email](https://react.email/docs/getting-started/manual-setup) docs suggest otherwise.

jirutka:
> Maybe, but the [react-email](https://react.email/docs/getting-started/manual-setup) docs suggest otherwise.

Well, the author of react-email either doesn’t know or doesn’t care about the difference between runtime and build-time/dev dependencies. :shrug: 

* https://www.geeksforgeeks.org/difference-between-dependencies-devdependencies-and-peerdependencies/
* https://frontendinterviewquestions.medium.com/difference-between-dependency-and-devdependency-2e8812b3f838

> To summarize, `dependencies` includes packages needed for the application to run in a production environment, while `devDependencies` includes packages required during development and testing but are not necessary for the production deployment.

Issue #1079, opened 2024-04-18
docker: reference to rallly_db should not contain port


tomquas:
hi. since rallly_db in docker-compose.yaml is a service reference, it cannot contain a port number 

actual:
```
- DATABASE_URL=postgres://postgres:postgres@rallly_db:5450/db
```

result:
```
Prisma schema loaded from prisma/schema.prisma
...
Error: P1001: Can't reach database server at `rallly_db`:`5450`
```

fix:
```
- DATABASE_URL=postgres://postgres:postgres@rallly_db/db
```

result:
```
Applying migration `..._init`
```

https://github.com/lukevella/rallly/blame/fc9149a3b2fa4401a8331b6d849066b504ea15e2/docker-compose.yml#L31

lukevella:
I believe this can be closed now.

PR #1132, opened 2024-05-29
Removed database port number from docker-compose.yml DATABASE_URL

dahlo:
## Description

As @tomquas pointed out in #1079, having the port number specified in the `DATABASE_URL` variable causes the `rallly_selfhosted` container not being able to connect to the `rallly_db` container.  `docker-compose up` fails with the message 

`Error: P1001: Can't reach database server at "rallly_db:5450"`

as it seems to think `:5450` is part of the hostname. Removing it solves the issue and the containers start as expected.

## Checklist

## Summary by CodeRabbit

- **Chores**
  - Updated database configuration to simplify the connection string by removing the port number.


lukevella:
Thanks @dahlo. I had to resolve some conflicts due to some recent changes but otherwise this looks good to me 👍

Issue #1007, opened 2024-01-29
Long names are not viewable


jahschwa:
**Describe the bug**
Long names are hidden by ellipsis, but there's no way to see the full name other than inspecting the raw HTML. If there is please correct me : )

Behavior appears to be coming from `.truncate` adding `text-overflow: ellipsis` to names.

**To Reproduce**

1. Use a very long name when answering a poll (I don't know the actual cutoff... I assume it's based on pixel length not character length, so let's just say 30 characters will probably always trigger it?)
2. View the poll; the name will be truncated with `...` but there's no way to see the full name

**Expected behavior**
Full name should be viewable somehow. Possibilities:
1. Add it as hover/alt text
2. Make participants list clickable and show a popup with all the full names
3. Reveal when clicking on it
4. Have a checkbox to toggle truncation
5. Add a scroll bar

**Screenshots**
![Screenshot from 2024-01-29 14-44-14](https://github.com/lukevella/rallly/assets/6052180/4eb024db-6d45-4302-90e5-4ef26740c405)

**Desktop (please complete the following information):**

- OS: Ubuntu 20.04
- Browser: Firefox 121.0.1
- Version: (I don't know what this field wants) https://app.rallly.co

aman-webdev:
Hi I also noticed this issue but in my case it doesnt truncate the name 
<img width="361" alt="image" src="https://github.com/lukevella/rallly/assets/78081991/cc59169d-8176-430d-a07e-17f9a4020455">
name used to test  :"This is a long name test and its working"

If its okay i can work on this issue

aman-webdev:
Also i noticed that we can save even if the name field is empty... I am not sure if this is a bug or expected output, it would be great if anyone can confirm

PR #1183, opened 2024-07-01
fix Long names are not viewable #1007

crnnr:
## Description

Fixed issue where long names would get cut off making them difficult/impossible to read by removing truncation by default

## Checklist

- [ x] I have performed a self-review of my code
- [ x] My code follows the code style of this project
- [ x] I have commented my code, particularly in hard-to-understand areas

## Summary by CodeRabbit

- **Style**
  - Updated the `UserAvatar` component to improve layout by removing unnecessary classes from the `<div>` element.


lukevella:
Removing `truncate` means that overflow is no longer handled. While truncating might not be the ideal solution it at least does not result in a seemingly broken UI. Unfortunately there is no obvious low-effort solution to this problem so some design considerations will not to be made to properly fix it.

PR #1515, opened 2025-01-26
♻️ Make user session optional

lukevella:
Rather than creating a guest in middleware, we now create a guest when a user creates a resource that needs to be tracked.


apps/web/src/components/user-provider.tsx
```diff
@@ -84,44 +83,45 @@ export const UserProvider = (props: { children?: React.ReactNode }) => {
...
+              timeZone: user.timeZone ?? null,
+              image: user.image ?? null,
+              locale: user.locale ?? i18n.language,
+            }
+          : {
+              id: "",
+              name: t("guest"),
+              isGuest: true,
+              tier: "guest",
+            },
```
lukevella, line 105: It might be better to update types so that `user` can be `undefined`.

PR #1541, opened 2025-02-03
Healthcheck for the docker

Billos:
## Description

Simple healthcheck to ensure the docker is healthy

## Checklist

- [x ] I have performed a self-review of my code
- [x ] My code follows the code style of this project
- [N/A ] I have commented my code, particularly in hard-to-understand areas

## Summary by CodeRabbit

- **New Features**
	- Integrated a proactive health monitoring feature that regularly verifies the application's responsiveness to enhance overall service reliability.
	- Added a health check command to verify application status.

### DIFF
--- a/apps/web/Dockerfile
+++ b/apps/web/Dockerfile
@@ -65,6 +65,6 @@ ARG SELF_HOSTED
 ENV NEXT_PUBLIC_SELF_HOSTED=$SELF_HOSTED
 ENV HOSTNAME=0.0.0.0
 
-HEALTHCHECK --interval=60s --timeout=20s curl --fail http://localhost:3000/api/auth/session || exit 1
+HEALTHCHECK CMD curl --fail http://localhost:3000/api/auth/session || exit 1
 
 CMD ["./docker-start.sh"]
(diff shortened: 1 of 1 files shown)


apps/web/Dockerfile
```diff
@@ -64,4 +65,6 @@ ARG SELF_HOSTED
 ENV NEXT_PUBLIC_SELF_HOSTED=$SELF_HOSTED
 ENV HOSTNAME=0.0.0.0
 
+HEALTHCHECK CMD curl --fail http://localhost:3000/api/auth/session || exit 1
```
lukevella, line 68: Some notes:
* We probably shouldn't be doing a healthcheck against the session endpoint since it's not its intended purpose.
* We should use the `$PORT` env var incase the value gets overridden.

End result would look something like this with the `api/status` endpoint still to be created.

```
HEALTHCHECK --interval=30s --timeout=5s --start-period=30s --retries=3 \
  CMD curl -f http://localhost:${PORT}/api/status || exit 1
```

Billos: Editing this !

Billos: That should be better. 
Once the api/status exists

lukevella:
LGTM. Thanks! 👍

PR #1643, opened 2025-03-27
✨ Update Auto-Delete Policy

lukevella:
Changes:
* Stop tracking last view
* Polls are auto-deleted when all dates are at least 30 days in the past

This change makes things a bit simpler, reduces the number of function calls being made and makes the `updated_at` column usable again since currently this column gets updated whenever we update the `touched_at` column automatically which isn't desirable.

## Summary by CodeRabbit

- **Documentation**
  - Updated FAQ content to clearly explain poll retention rules: free users’ polls are marked for deletion when options are older than 30 days, while Pro users’ polls are preserved.

- **Refactor and Cleanup**
  - Revised poll cleanup logic to rely solely on option dates.
  - Increased batch processing for poll deletion, now targeting polls marked as deleted for over 30 days.
  - Removed `useTouchBeacon` functionality from relevant components.

- **Deprecations**
  - Deprecated legacy poll update methods and fields in preparation for future improvements.


lukevella:
One concern about this is that a large number of polls will be marked for deletion where in the past they have been kept alive due to visitors resetting the auto-deletion timer. We can assume that a significant amount of users is going to be seeing 404 pages instead of the poll they are expecting which does not make for a great user experience. For some users, it may even be their first time using Rallly.

It might be worth considering that we should handle this change a bit more gracefully for the sake of our existing user-base. Perhaps we can keep the same policy and leverage our kv store as a way to keep track of whether a poll has been visited or not. Or have a separate table to track visits. This will handy if we would like to offer analytics (for example).

Issue #1218, opened 2024-07-22
Timezone changes when all options are removed using "edit options" path


jonnymarshall:
**Describe the bug**
When you go to "edit options", the timecode of the poll is preserved unless you delete ALL the existing options and then begin adding new ones, at which point the timezone defaults to your current location and cannot be changed.

**To Reproduce**

1. Go to an existing poll with a timezone different to your current timezone
2. Delete all the poll options
3. Add a new option
4. The new option will have changed to your current timezone and be uneditable.

**Expected behavior**
Whether you delete some OR all of the poll options, the timezone of the poll should be preserved.

- OS: Mac OS
- Browser: Chrome
- Version: 126.0.6478.183

lukevella:
Thanks for reporting and for the reproduction steps. Working on a fix.

lukevella:
Reopening this since the fix had to be reverted

cdznd:
Hello @lukevella, I was looking for ways to contribute to this project and came across this bug. I was able to reproduce it as described in the issue and after some debugging, I found a reasonable solution. Can I work on it and have it assigned to me?

PR #1659, opened 2025-04-07
🐛 Fix automatic timezone change on poll update

cdznd:
## Description

**Bug Fix**
Fixes issue: [#1218](https://github.com/lukevella/rallly/issues/1218)
Fixes a bug where the timezone resets to the browser’s timezone when all poll options are removed using the "Edit Options" path or when toggling the "Specify times" option off and on again.

Previously, when all time options were deleted from a poll or when the "Specify times" option was disabled and re-enabled, the timezone would revert to the user's current (browser) timezone. This PR ensures that the originally selected timezone is preserved in these cases.

### QA Instructions:

**On Poll Creation**:
1. Enable the "Specify times" option.
2. Verify that the browser’s timezone is shown by default.
3. Select a different timezone.
4. Disable and re-enable the "Specify times" option — the selected timezone should remain unchanged.
5. With "Specify times" disabled, go to week view and add a new time — the selected timezone should still be preserved.
6. Toggling the "Automatic Time Zone Conversion" option should reset the timezone to the browser’s timezone (this is expected behavior).

**On Poll Update**:
1. Open an existing poll with a timezone different from your current browser timezone.
2. Delete all poll options.
3. Add a new option.
4. Enable the "Specify times" option.
5. The originally selected timezone should be preserved and remain editable.

## Checklist

- [ ] I have performed a self-review of my code
- [ ] I have commented my code, particularly in hard-to-understand areas

## Summary by CodeRabbit

- **New Features**
  - Improved time zone management in the poll scheduling form. When toggling the time specification option, the interface now preserves your previously selected time zone for a more consistent experience.

### DIFF
--- a/apps/web/src/components/forms/poll-options-form/month-calendar/month-calendar.tsx
+++ b/apps/web/src/components/forms/poll-options-form/month-calendar/month-calendar.tsx
@@ -52,6 +52,8 @@ const MonthCalendar: React.FunctionComponent<DateTimePickerProps> = ({
 
   const form = useFormContext<NewEventData>();
 
+  const [preservedTimeZone, setPreservedTimeZone] = React.useState<string | null>(null);
+
   const optionsByDay = React.useMemo(() => {
     const res: Record<
       string,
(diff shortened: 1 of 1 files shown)


lukevella:
Hi @cdznd, thanks for taking the time to look into this. It's been a while since I looked at this issue, but having tried to fix it myself (#1221) and then having to revert it (#1227) I remember finding that there were some deeper issues here that made it difficult to fix completely without introducing other bugs. I'm currently focusing on a major update and plan to replace the entire implementation so I can't justify spending the time to review this PR.

PR #2436, opened 2026-06-15
🐛 Allow changing the time zone on polls without votes

### LINKED ISSUE #1085: Bug: Cannot change timezone after an event has been created.
**Describe the bug**
Cannot change timezone after an event has been created.

**To Reproduce**
1. Create an event
2. Go to 'manage' > 'edit options'
3. Attempt to click on where it tells you the timezone, or 'Automatic Time Conversion' (see screenshot)
4. Neither option is clickable. The former is greyed out. The latter changes the cursor to the unactionable symbol.

**Expected behavior**
You should be able to click and change the timezone on the right. And toggle 'Automatic Time Conversion' on/off after creating an event.

**Screenshots**
Screenshot 2024-05-01 at 09 45 50

**Desktop (please complete the following information):**

- OS: Mac OS Sonoma Version 14.4.1 (23E224)
- Browser: Chrome 
- Version: Version 124.0.6367.92 (Official Build) (arm64)

**Additional context**
I seem to remember that in the past this feature has worked. But right now I can't change either of these settings on either a new or past scheduling poll.

crossi-dev:
Fixes #1085.

**Problem:** the timezone selector and auto-detect toggle are locked once a poll is created — you can't change the timezone after creation.

**Why it was locked:** in #1037 the controls were disabled unconditionally to prevent a data-safety issue — changing the timezone re-runs `convertOptionToString` with the new TZ, making every existing option look "changed", so options get deleted/recreated and all votes are wiped.

**Fix:** make the guard *conditional* — `disableTimeZoneChange = hasVotes` (computed from the already-fetched participants). Polls with **zero votes** have nothing to corrupt, so the controls are fully editable again. When votes exist, the controls stay locked and the tooltip now explains why ("Time zone cannot be changed after votes have been cast. To change it, ask participants to clear their votes first."), which also addresses the UX confusion in the report.

**Verified:** `tsc --noEmit` clean on the changed files; `timezone-change-detector` (5/5) and `time-zone-select` (8/8) tests pass.

---

Came across rallly and noticed #1085 had been open a while, so I fixed it — scoped to the zero-votes case so it doesn't reintroduce the vote-reset issue from #1037. No strings. I do this kind of work (Choreless — we fix and ship code for small teams), so if you ever want a hand, I'm around. — Charles

## Summary by CodeRabbit

* **New Features**
  * Event time zone changes are now available only before any votes are submitted.
  * After voting begins, time zone changes are disabled to prevent inconsistent scheduling.
* **Documentation**
  * Updated the event time zone help/tooltip text to explain why the time zone can’t be changed once votes are cast.

### DIFF
--- a/apps/web/public/locales/en/app.json
+++ b/apps/web/public/locales/en/app.json
@@ -64,7 +64,7 @@
   "authErrorsUserBanned": "This account has been banned. Please contact support if you believe this is an error.",
   "authErrorsUserNotFound": "No account found with this email address. Please check the email or register for a new account.",
   "autoTimeZone": "Automatic Time Zone Conversion",
-  "autoTimeZoneDisabledHelp": "Time zone cannot be changed after votes have been cast. To change the time zone, ask participants to clear their votes first.",
+  "autoTimeZoneDisabledHelp": "The time zone can't be changed after votes have been cast.",
   "autoTimeZoneHelp": "Enable this setting to automatically adjust event times to each participant's local time zone.",
   "back": "Back",
   "banned": "Banned",
--- a/apps/web/public/locales/en/app.json
+++ b/apps/web/public/locales/en/app.json
@@ -64,6 +64,7 @@
   "authErrorsUserBanned": "This account has been banned. Please contact support if you believe this is an error.",
   "authErrorsUserNotFound": "No account found with this email address. Please check the email or register for a new account.",
   "autoTimeZone": "Automatic Time Zone Conversion",
+  "autoTimeZoneDisabledHelp": "Time zone cannot be changed after votes have been cast. To change the time zone, ask participants to clear their votes first.",
   "autoTimeZoneHelp": "Enable this setting to automatically adjust event times to each participant's local time zone.",
   "back": "Back",
   "banned": "Banned",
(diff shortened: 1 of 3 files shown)


lukevella:
Hi @crossi-dev, thanks for taking the time to look into this and for the clear write-up. The conditional guard on `hasVotes` is a reasonable approach to #1085.

That said, I'm going to close this without merging. I'm actively refactoring the poll creation and edit flow right now, and the code this PR touches (including `poll-options-form`) is going to change significantly as part of that work. Merging changes into this area at the moment would add churn for both of us, since the fix would likely be rewritten or dropped during the refactor.

The underlying issue in #1085 is still valid and I'll make sure the new flow handles timezone changes properly, taking the vote-safety concern from #1037 into account.

Thanks again for the contribution.

lukevella:
@crossi-dev apologies for backtracking here. After giving it more thought, I've decided this is actually a reasonable solution for the short term while we work on replacing this flow, so I'm reopening it.

Since you opened this PR, a couple of refactors landed on main that conflict with the branch. Maintainer edits were enabled, so I've pushed the rebase and adaptations myself — see the comment below for details.

lukevella:
Update: maintainer edits were already enabled on this branch, so I went ahead and pushed the updates myself rather than wait. Your commit is preserved as is, rebased onto main, with a second commit on top that adapts it to the refactors that landed since June:

- the form moved to `apps/web/src/features/poll/components/forms/poll-options-form/`
- the tooltip wrapper around `TimeZoneSelect` was removed since Base UI's `TooltipTrigger` has no `asChild`; the info icon tooltip now carries the explanation when the picker is locked
- the disabled message is now "The time zone can't be changed after votes have been cast."
- the translation key is generated via `pnpm i18n:scan`

Heads up that this required a force push since the branch needed a rebase, so you'll want to reset your local copy if you still have one. Thanks again for the contribution.

PR #2442, opened 2026-06-16
🗃️ Enforce one invite per email on a scheduled event

lukevella:
## Why

Follow-up to #2441. Scheduled events should hold at most one invite per email, but nothing enforced it: a poll can have several participants sharing an email, and finalizing it created one invite per participant. The public RSVP path also used a racy check-then-create, so concurrent submissions could duplicate. (Resolves the duplicate-RSVP race CodeRabbit flagged in #2441.)

## What

- **Unique index** `(scheduled_event_id, invitee_email)` on `scheduled_event_invites`. `invitee_email` is `citext`, so it's case-insensitive.
- **Backfill in the migration, before the index.** Collapses existing duplicates, keeping one per `(event, email)` by preferring an account-linked invite, then the most committal status (`accepted > tentative > declined > pending`), then most recent activity. Measured against prod: 45 groups / 55 rows, **0 groups with 2+ distinct real accounts** (so no genuine identity collisions), 22 same-person status divergences.
- **Finalize (`polls.ts`)** dedupes same-email participants into a single invite, keeping the most committal status.
- **Public RSVP (`mutations.ts`)** now relies on the constraint (catch `P2002`) instead of a racy precheck.

## Deploy note

The migration **deletes 55 duplicate invite rows** in prod before creating the index. The keep-rule is in the migration SQL — worth an eyeball before deploy. Postgres can't build the unique index while duplicates exist, so the delete and the index are in one migration.

## Out of scope

Poll-participant uniqueness at source (the "verify email to manage your poll response" feature) is a separate initiative; this PR only constrains the invite table.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of duplicate event invitations to prevent errors when users accidentally respond multiple times.
  * Enhanced poll scheduling to properly consolidate multiple participants sharing the same email address into a single invite.


lukevella:
@copilot resolve the merge conflicts in this pull request

PR #2464, opened 2026-06-19
Migrate database to Supabase (isolated rallly schema)

amateurnerd44:
## Why
Production (`rally.wanakusuite.com`) was returning HTTP 500 on every request. Root cause: the old GCP Postgres at `34.63.8.148` became unreachable when the GCP free trial expired. Every server render calls Prisma (branding/license, user/session, billing tier), so each one threw `P1001: Can't reach database server`.

## What
Point the app at the shared suite Supabase project (`xkzwyatepjhwmavlqzqv`). To avoid colliding with the portal's tables in that shared project (it owns `public.users/accounts/sessions`), Rallly's ~28 tables live in a dedicated **`rallly` Postgres schema**.

- **`client.ts`** — the `pg` driver adapter now targets `DATABASE_SCHEMA` when set, so runtime queries hit the `rallly` schema. Unset locally → default `public`.
- **`env.ts`** — validate `DIRECT_DATABASE_URL` (session connection for Prisma CLI/migrations) and `DATABASE_SCHEMA`.

## Already done out of band
- `rallly` schema created; `prisma migrate deploy` applied all 114 migrations there (verified: 28 app tables in `rallly`, portal's `public` tables untouched).
- Vercel env set (Production + this branch's Preview): `DATABASE_URL` (6543 transaction pooler), `DIRECT_DATABASE_URL` (5432 session, `?schema=rallly`), `DATABASE_SCHEMA=rallly`.
- End-to-end smoke test: the three root-render queries succeed via the 6543 pooler against the `rallly` schema.

## Release Notes

* **New Features**
  * Introduced "Continue with RepSuite" login option for streamlined authentication through the Wanaku Suite portal
  * Enabled automatic sign-in when RepSuite authentication is configured
  * Added Wanaku Suite portal navigation link in the app sidebar

* **Behavior Changes**
  * Sign-out now performs suite-wide logout through the portal instead of local-only logout


amateurnerd44:
Opened against the wrong repo (fork parent). Closing.

PR #2470, opened 2026-06-19
🐛 Keep all-day poll options on the same date across timezones

### LINKED ISSUE #2467: Creating all-day polls becomes confusing to users in earlier timezones
**Describe the bug**
When a Poll is created in f.ex. UTC+1 with an entire day marked as option (f.ex. 2nd of july) it will display as the previous day for users in earlier timezones (so a user in UTC+0 would see 1st of july) without any clear warning about this.

This also affects browsers that for finger printing reason tell the service they are in UTC+0 (which is how I noticed it).

**To Reproduce**
1. Create a poll in UTC+1 and set one or multiple days as option
2. Open the poll with a user that has UTC+0 as preferred timezone

**Expected behavior**
Either day schedules should not adjust to timezones (because you are selecting the day and not a time)
Or there should be a clear warning that you are currently responding to a poll in a different timezone and dates may be changed.
Ideally with an option to switch to the timezone of the poll

**Screenshots**
Browser with fingerprinting protection:
Image

Other browser:
Image

**Desktop (please complete the following information):**

- OS: Linux Mint
- Browser with fingerprinting protection: Librewolf
- Other Browser: Thorium
- site: rallly.co

lukevella:
Fixes #2467

## Problem

All-day (date-only) poll options were stored against the creator's timezone and then converted into each viewer's timezone on display. A date picked as **Jul 1** rendered as **Jun 30** for participants in an earlier timezone — the day-of-week shifted too. Reported on rallly.co with screenshots showing the same poll on two browsers a day apart.

An all-day option is a *floating date*: "July 1" everywhere on Earth. It must never be timezone-converted.

## Root cause

A date-only poll should carry **no timezone** — a falsy `poll.timeZone` is meant to be the single source of truth for "this option is floating". But the create flow attached the browser timezone to date polls anyway (~6.3k polls in prod), so every display site dutifully converted their dates to the viewer's zone. The option timestamps were also stored with the offset baked in (`Jul 1` in UTC+1 → `2026-06-30 23:00Z`) instead of UTC midnight.

## Fix

Enforce the invariant at the data layer instead of special-casing every display site:

- **`make` / `edit` mutations** drop the timezone for date-only polls (`isTimePoll ? input.timeZone : null`). Date options are then stored at UTC midnight and can never shift. Time polls are unchanged.
- **Poll grid** reads date options in UTC directly — defense in depth in the one function that already classifies date vs time options.
- The finalize dialog, scheduled-event banner, and finalize storage are **unchanged** — they already trust `poll.timeZone`, which is now reliably null for date polls.

## Data migration

`20260619150228_fix_all_day_option_timezone_shift` repairs existing data, in order:

1. Snap offset-baked all-day **option** timestamps to UTC midnight of the creator's intended date (using the old timezone — must run before step 3).
2. Same for poll-linked all-day **scheduled events** (`start`/`end`).
3. Null `time_zone` on date-only polls (has all-day options, no timed options).

Guards make it idempotent (`::time <> '00:00:00'` skips already-correct rows) and abort-proof (`pg_timezone_names` filter skips unparseable timezones without erroring). Verified against Postgres 18 across DST, fractional (+5:30, +5:45), negative, and far-positive offsets, plus a poison invalid-tz row.

### ⚠️ Pre-deploy steps (migration is irreversible)

1. **Snapshot the database first.** The migration overwrites `start_time` and nulls `time_zone` with no record of the originals.
2. **Run this diagnostic on prod** to confirm every affected poll uses a standard IANA zone:
   ```sql
   SELECT p.time_zone, count(DISTINCT p.id) AS polls
   FROM polls p
   JOIN options o ON o.poll_id = p.id
   WHERE o.duration_minutes = 0
     AND p.time_zone IS NOT NULL
     AND p.time_zone NOT LIKE '%/%'
   GROUP BY p.time_zone ORDER BY polls DESC;
   ```
   Empty (or only `UTC`/`GMT`) → clean, proceed. Anything else (`EST`, junk) → review first; abbreviation zones snap with a fixed offset and non-IANA junk is skipped.

**Out of scope:** mixed polls (both all-day and timed options) are forbidden by the form; the migration conservatively keeps their timezone and neither fixes nor worsens their pre-existing display behavior.

## Testing

- New regression test `poll-context.test.ts` — all-day options stay on the same date across source/target timezones (fails without the fix).
- `pnpm type-check`, `pnpm check`, and existing `ics.test.ts` pass.
- Migration SQL validated against the live Postgres 18 schema in a rolled-back transaction.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed all-day polls to display consistently across different timezones. All-day options now render as their intended calendar date regardless of the viewer's timezone or poll creator's timezone settings.

### DIFF
--- a/apps/web/src/components/poll-context.test.ts
+++ b/apps/web/src/components/poll-context.test.ts
@@ -1,0 +1,46 @@
+import { describe, expect, it } from "vitest";
+
+import { createOptionsContextValue } from "./poll-context";
+
+// An all-day option is a floating date: "July 1" everywhere on Earth. It must
+// render as the same calendar date regardless of the poll timezone (source) or
+// the viewer timezone (target). Regression test for the bug where date options
+// were timezone-shifted, so a viewer in an earlier timezone saw the prior day.
+describe("createOptionsContextValue — all-day options are timezone-invariant", () => {
+  const allDayOption = {
+    id: "opt-1",
+    startTime: new Date("2026-07-01T00:00:00Z"),
+    duration: 0,
+  };
+
+  it.each([
+    ["America/Los_Angeles", "Europe/Berlin"], // viewer far behind source
+    ["Europe/Berlin", "America/Los_Angeles"], // viewer far ahead of source
+    ["Asia/Tokyo", "Europe/Berlin"],
+    ["UTC", "Europe/Berlin"],
+  ])("renders Jul 1 for target=%s, source=%s", (targetTimeZone, sourceTimeZone) => {
+    const result = createOptionsContextValue(
+      [allDayOption],
+      targetTimeZone,
+      sourceTimeZone,
+    );
+
+    expect(result.pollType).toBe("date");
+    expect(result.options[0]).toMatchObject({
+      type: "date",
+      day: "1",
+      month: "Jul",
+      year: "2026",
+    });
+  });
+
+  it("does not shift when the poll has no timezone", () => {
+    const result = createOptionsContextValue(
+      [allDayOption],
+      "America/Los_Angeles",
+      null,
+    );
+
+    expect(result.options[0]).toMatchObject({ day: "1", month: "Jul" });
+  });
+});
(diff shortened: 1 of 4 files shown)


lukevella:
@coderabbitai review

PR #2482, opened 2026-06-22
🧹 Add script to deactivate stale seat-update portal configs

lukevella:
## Context

Follow-up to #2481. `createQuantityUpdatePortalConfiguration()` ([apps/web/src/features/billing/utils.ts](apps/web/src/features/billing/utils.ts)) creates a **new** Stripe billing portal configuration on every seat-update session and never reuses it, so the account slowly accumulates throwaway config objects.

## What this adds

A dry-run-by-default script [`packages/billing/src/scripts/cleanup-portal-configurations.ts`](packages/billing/src/scripts/cleanup-portal-configurations.ts) that finds the stale seat-update configs and deactivates them.

```bash
# Dry run — lists what would be deactivated; changes nothing
pnpm --filter @rallly/billing cleanup-portal-configurations

# Deactivate them
pnpm --filter @rallly/billing cleanup-portal-configurations -- --apply
```

## Important caveats

- **Deactivate, not delete.** Stripe has **no delete endpoint** for billing portal configurations. The only cleanup possible is `active: false` — the objects remain in the account, just inactive. The script and its naming reflect this.
- **Targeted matching.** Only configs whose `business_profile.headline` equals `"Update your seat allocation"` are touched, and the default config is always skipped — so a manually-configured or default portal can't be affected. The headline constant must stay in sync with the one in `utils.ts`.
- **This is hygiene, not a fix.** Leaving the configs causes no billing or functional harm. The actual root-cause fix is to cache/reuse a single configuration instead of creating one per session — a separate change, not done here.

## Summary by CodeRabbit

* **Chores**
  * Added a new `cleanup-portal-configurations` script for managing Stripe Billing Portal configurations, which runs in dry-run mode by default and supports an `--apply` flag to execute permanent changes.

### DIFF
--- a/packages/billing/package.json
+++ b/packages/billing/package.json
@@ -8,6 +8,7 @@
   },
   "scripts": {
     "checkout-expiry": "dotenv -e .env -- pnpx tsx ./src/scripts/checkout-expiry.ts",
+    "cleanup-portal-configurations": "dotenv -e .env -- pnpx tsx ./src/scripts/cleanup-portal-configurations.ts",
     "invoice-pending-prorations": "dotenv -e .env -- pnpx tsx ./src/scripts/invoice-pending-prorations.ts",
     "subscription-data-sync": "dotenv -e .env -- pnpx tsx ./src/scripts/subscription-data-sync.ts",
     "sync-payment-methods": "dotenv -e .env -- pnpx tsx ./src/scripts/sync-payment-methods.ts",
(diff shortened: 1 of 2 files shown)


lukevella:
Superseded by the consolidated PR that makes the cleanup script version-aware and adds the reuse fix. A headline-only deactivator is unsafe once a versioned singleton config exists (it would deactivate the live config), so folding both into one PR.

PR #2529, opened 2026-07-06
♻️ Migrate license page purchase date to Intl

lukevella:
Part of RAL-1247.

Replaces the last user-facing dayjs format string in the control panel: the license purchase date now renders with `formatDate` from `lib/datetime` in the request locale (`Jul 6, 2026` style instead of hardcoded `YYYY-MM-DD`). All-day/UTC semantics apply since it's a date-only value.

## Summary by CodeRabbit

* **New Features**
  * Dates and times in the control panel now consistently follow your account settings, and align more reliably with your active locale across the app.
* **Bug Fixes**
  * Improved the license “Purchase Date” display to use the correct account date formatting for clearer, consistent results.
* **Refactor**
  * License and date/time information is loaded before rendering to ensure the correct values are shown immediately.

### DIFF
--- a/apps/web/src/app/[locale]/control-panel/license/page.tsx
+++ b/apps/web/src/app/[locale]/control-panel/license/page.tsx
@@ -43,19 +43,14 @@ import { RemoveLicenseButton } from "@/features/licensing/components/remove-lice
 import { loadInstanceLicense } from "@/features/licensing/data";
 import { Trans } from "@/i18n/client";
 import { getTranslation } from "@/i18n/server";
-import { formatDate } from "@/lib/datetime/format";
+import { Time } from "@/lib/datetime/time";
 
 async function loadData() {
   const license = await loadInstanceLicense();
   return { license };
 }
 
-export default async function LicensePage({
-  params,
-}: {
-  params: Promise<{ locale: string }>;
-}) {
-  const { locale } = await params;
+export default async function LicensePage() {
   const { license } = await loadData();
 
   return (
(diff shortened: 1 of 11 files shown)


apps/web/src/app/[locale]/control-panel/license/page.tsx
```diff
@@ -115,7 +120,7 @@ export default async function LicensePage() {
                     <Trans i18nKey="purchaseDate" defaults="Purchase Date" />
                   </DescriptionListTitle>
                   <DescriptionListValue>
-                    {dayjs(license.issuedAt).format("YYYY-MM-DD")}
+                    {formatDate(license.issuedAt, { locale, preset: "date" })}
```
Comment (lukevella) [inline apps/web/src/app/[locale]/control-panel/license/page.tsx]: Both the control panel and the space layout are accessible only by authenticated users. Date formatting in these areas should adhere to the user's locale and date preferences. It would make sense to wrap them both in shared layout with the context provider that allows us to use components and hooks that don't require forwarding these preferences.

However to avoid a big restructuring in this PR, we can just wrap the control panel layout with a `DateTimeProvider`.

Comment (lukevella) [reply apps/web/src/app/[locale]/control-panel/license/page.tsx]: Done in 9980e5256. The control panel layout now wraps its tree in `DateTimeProvider` seeded from `user.getAuthed` (`timeZone` / `timeFormat` / `weekStart`, same shape as the (space) layout), and the license page renders the purchase date with the context-bound `<Time preset="date" />` instead of forwarding the locale explicitly. The page no longer needs `params`, and the date now respects the admin's zone and format preferences rather than forcing UTC.

Comment (lukevella) [reply apps/web/src/app/[locale]/control-panel/license/page.tsx]: You're right — done in 3f3eff907. `locale` is now a required prop on `DateTimeProvider` (and `DeviceDateTimeProvider`), and the `useParams` fallback in `useDateTimeConfig` is gone, so nothing depends on the locale URL segment anymore. All seven provider call sites (control panel, (space), (auth), (optional-space), invite, event, quick create) now resolve it server-side via `getLocale()` and forward it explicitly.

PR #2530, opened 2026-07-06
♻️ Migrate event email datetime formatting to Intl

lukevella:
Part of RAL-1247 — the last consumer migration: `formatEventDateTime` now builds the finalized/canceled email strings with the pure `lib/datetime` functions instead of hardcoded English dayjs patterns.

## Changes

- **`formatEventDateTime`** takes the recipient's `locale` (default `en`, matching the email templates' fallback) and formats with `formatDateTime` / `formatDateTimeRange` / `formatDateParts`. Zone semantics, now documented on the function:
  - all-day → UTC, same date for everyone
  - fixed (has a zone) → `inviteeTimeZone ?? event.timeZone`, zone name appended (`9:00 – 10:00 AM EDT`)
  - floating (no zone) → UTC wall time, no zone name
- **All-day label moved into the templates**: `time` is now optional; when absent, the finalized-host / finalized-participant / event-canceled templates render `t("allDay")` via their own i18n, so it localizes with the rest of the email instead of the hardcoded English "All day".
- **Cancellation emails now use `inviteeLocale`** — it was stored but never passed, so those emails always rendered in English and the old formatter's output was English regardless of recipient.
- Unit tests pin all four zone scenarios plus locale output.

## Behavior notes

- Times now follow the recipient locale's hour cycle (en → 12h) instead of hardcoded `HH:mm`.
- `Intl.DateTimeFormat.formatRange` adds the date when a range crosses midnight in the display zone, removing the old ambiguous `18:00 - 00:00` output.

## Summary by CodeRabbit

* **New Features**
  * Event date/time formatting now respects each recipient’s locale (including 24-hour `timeFormat` when set).
  * All-day events are shown as a localized “All day” in event views and email templates.
* **Bug Fixes**
  * Cancelation and finalization emails now use matching locale/time formatting for hosts and invitees.
  * Time zone display and fallback behavior are more consistent for UTC, fixed-time, and floating events.
* **Tests**
  * Added formatting tests covering all-day, fixed-time, floating, time-zone, 24-hour, and localized scenarios.

### DIFF
--- a/apps/web/src/trpc/routers/polls.ts
+++ b/apps/web/src/trpc/routers/polls.ts
@@ -966,16 +966,18 @@ export const polls = router({
           });
         }
 
+        const hostEmail = poll.user.email;
+        const hostName = poll.user.name;
+        const hostLocale = poll.user.locale ?? undefined;
+
         const { date, day, dow, time } = formatEventDateTime({
           start: scheduledEvent.start,
           end: scheduledEvent.end,
           allDay: scheduledEvent.allDay,
           timeZone: scheduledEvent.timeZone,
+          locale: hostLocale,
         });
 
-        const hostEmail = poll.user.email;
-        const hostName = poll.user.name;
-        const hostLocale = poll.user.locale ?? undefined;
         const space = poll.space;
         after(async () =>
           sendFinalizeHostEmail({
@@ -1014,6 +1016,7 @@ export const polls = router({
             allDay: scheduledEvent.allDay,
             timeZone: scheduledEvent.timeZone,
             inviteeTimeZone: p.timeZone,
+            locale: p.locale,
           });
           after(async () =>
             sendFinalizeParticipantEmail({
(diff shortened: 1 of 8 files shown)


apps/web/src/trpc/routers/polls.ts
```diff
@@ -966,16 +966,18 @@ export const polls = router({
...
+        const hostEmail = poll.user.email;
+        const hostName = poll.user.name;
+        const hostLocale = poll.user.locale ?? undefined;
+
         const { date, day, dow, time } = formatEventDateTime({
           start: scheduledEvent.start,
           end: scheduledEvent.end,
           allDay: scheduledEvent.allDay,
           timeZone: scheduledEvent.timeZone,
+          locale: hostLocale,
```
lukevella, line 979: Can we set the host's time format as well?

lukevella: Done in 3a3e26ab0. `formatEventDateTime` takes an optional `timeFormat`, the book query selects `user.timeFormat`, and the host email call passes it — so the host sees their preferred hour cycle instead of the locale default.

Left participants and invitees on the locale default: their records store `locale`/`timeZone` but no time format, and most aren't registered. Registered participants do have a preference on their user row (`p.user.timeFormat` is one select away) — say the word if you want that threaded through too.

PR #2545, opened 2026-07-08
✨ Add maintenance mode

lukevella:
## Summary

Adds a maintenance mode so we can take the app offline for scheduled maintenance, such as migrating to a new database. Design discussion and plan: [RAL-1275](https://linear.app/rallly/issue/RAL-1275/maintenance-mode).

Maintenance mode is an availability feature, not a security boundary — enforcement is deliberately lightweight:

- **Toggle**: `MAINTENANCE_MODE=true` env var (cloud: flip in Vercel + redeploy; self-hosted: set env + restart).
- **Pages**: the proxy redirects all page traffic to a new `/maintenance` page. When maintenance is off, `/maintenance` redirects home.
- **Execution guards**: the tRPC base procedure, the base safe-action client, and the private Hono API reject with `SERVICE_UNAVAILABLE` (503) — this prevents stray writes to the old database from in-flight tabs and API clients during a migration window, instead of letting requests fail raw against a down database. Maintenance 503s are excluded from Sentry capture so a window doesn't flood it.
- **Operator bypass**: with `MAINTENANCE_BYPASS_TOKEN` set, visiting `/maintenance?token=<secret>` sets a 4h httpOnly cookie (sha256 of the token, compared timing-safe via Web Crypto since the proxy runs on the edge runtime) that lets you fully use the app to verify it against the new database before reopening traffic.
- **Keeps working with the DB down**: `/api/status` is untouched (Docker HEALTHCHECK), `docker-start.sh` skips `prisma migrate deploy` when maintenance is on so the container can boot, and the self-hosted license lookup in the branding query is skipped so the maintenance page renders without a database.

## Verification

Drove the running app in all three configurations (unit tests + type-check + Biome also pass):

| Scenario | Result |
|---|---|
| Maintenance on: `GET /` | 307 → `/maintenance`, branded page renders |
| Maintenance on: `GET /api/status` | 200 |
| Maintenance on: real tRPC procedure | 503 `SERVICE_UNAVAILABLE` |
| Maintenance on: `GET /api/private/*` | 503 JSON + `Retry-After: 300` |
| Correct bypass token | 307 → `/` + hashed httpOnly cookie; pages **and** tRPC fully usable |
| Wrong token / no token configured | maintenance page, no cookie, no oracle |
| Stale bypass cookie after secret removed | rejected, redirected to `/maintenance` |
| Maintenance off | `/`, `/login`, tRPC all normal; `/maintenance` → `/` |

## Runbook (cloud DB migration)

1. Set `MAINTENANCE_MODE=true` (+ bypass token) and redeploy **while the old DB is still reachable** (the build runs `db:deploy`)
2. Migrate the data
3. Point `DATABASE_URL` at the new database, verify via the bypass URL
4. Set `MAINTENANCE_MODE=false` and redeploy

Fixes RAL-1275

## Summary by CodeRabbit

* **New Features**
  * Added maintenance mode support across the app, including a dedicated maintenance page and localized downtime messaging.
  * Visitors can now be redirected to a maintenance page, with an optional bypass token to regain access temporarily.

* **Bug Fixes**
  * APIs and server actions now return a clear “service unavailable” response during maintenance.
  * Startup behavior now skips database migrations when the database is unreachable in maintenance mode.

* **Documentation**
  * Added setup and usage details for maintenance mode and bypass access.


apps/docs/self-hosting/configuration.mdx
```diff
@@ -182,6 +182,20 @@ The stack bundles [Garage](https://garagehq.deuxfleurs.fr/) for file uploads, wh
   Secret key for the bucket.
 </ParamField>
 
+### Maintenance mode
```
lukevella: It only makes sense to publish this after we make a new release and we should be clear what release it requires to use these new flags.

lukevella: Removed in 62925d64c — the docs will be added once 4.12 is released, tracked in [RAL-1278](https://linear.app/rallly/issue/RAL-1278/document-maintenance-mode-in-self-hosted-configuration-docs) with the content ready to paste (including "Available from v4.12.0" notes).

PR #2552, opened 2026-07-09
🗃️ Backfill all-day event encoding and enforce invariants with CHECKs

lukevella:
Steps 1 and 3 of the RAL-1257 migration plan. Stacked on #2551 (the write fixes) — **must not merge or deploy before it**, otherwise the old finalize code can write rows the constraints reject.

## What it does

One raw SQL migration, backfill first, constraints second:

1. **Zoned all-day rows already at UTC midnight** — only null the zone, never shift the date (the trap flagged in the audit; 2 rows in prod, including `ny4Tl8CBmScc` which currently displays a day early in the event list). What invitees received by email doesn't change.
2. **Zoned all-day rows at zone midnight** — snap to the zone's calendar date (0 rows in prod; guard for self-hosted databases).
3. **Zero-length all-day rows** from the end date bug fixed in #1926 — restore the 1 day span (40 rows in prod).
4. **Catch-all**: any remaining all-day row off UTC midnight snaps to the UTC calendar date (0 rows in prod) so the constraints can never fail the migration on a self-hosted database.
5. Adds `all_day_is_floating` and `all_day_is_utc_midnight` CHECK constraints.

The catch-all (4) is an addition over the issue's plan: without it, an unexpected row in a self-hosted database would abort the migration and block startup.

## Testing

Verified against a scratch database on the dev postgres container:

- Seeded one fixture row per backfill branch plus correct all-day and timed rows; deployed the migration; every branch produced the expected result and correct rows were untouched (timed rows keep their zone)
- Post-migration inserts violating each invariant (zoned all-day, off-midnight all-day, zero-length all-day) are rejected by the matching constraint; valid all-day and timed inserts succeed

## Summary by CodeRabbit

* **New Features**
  * Poll scheduling now handles timed and all-day options more consistently across time zones.
  * Added improved event time handling for scheduled poll outcomes.

* **Bug Fixes**
  * Fixed mixed polls so date-only options are stored as floating UTC dates instead of being shifted by the poll time zone.
  * Corrected all-day event behavior to ensure proper start/end dates and remove invalid time zone values.
  * Existing scheduled events were normalized to prevent zero-length or misaligned all-day entries.

### DIFF
--- a/apps/web/src/trpc/routers/polls.ts
+++ b/apps/web/src/trpc/routers/polls.ts
@@ -30,6 +30,7 @@ import {
 } from "../trpc";
 import { comments } from "./polls/comments";
 import { participants } from "./polls/participants";
+import { getScheduledEventTimes } from "./polls/scheduled-event-times";
 import { timeZoneInput } from "./polls/schema";
 
 const collapseNewlines = (s: string) => s.replace(/\n{3,}/g, "\n\n");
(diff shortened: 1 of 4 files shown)


lukevella:
@coderabbitai review

PR #2598, opened 2026-07-10
🐛 Make redirectIfLoggedIn bypass the session cookie cache

lukevella:
## Summary

Part 1 of [RAL-1313](https://linear.app/rallly/issue/RAL-1313/fix-infinite-redirect-loop-between-login-and-for-stalebanned-sessions): PostHog shows users bouncing between `/login` and `/` in an infinite loop. One leg of that loop is `redirectIfLoggedIn`, which trusted the 5 minute session cookie cache and ignored ban status, so `/login` kept redirecting banned or revoked sessions to `/` where they immediately fail and get pushed back.

This makes `/login` (and `/register`, `/forgot-password`, `/login/verify`) authoritative:

- Calls `authLib.api.getSession` with `disableCookieCache: true` so the redirect decision is backed by a DB read instead of a possibly stale cookie cache.
- Only redirects when the user is non-anonymous **and** not banned.
- A failed session lookup logs and renders the page instead of throwing, since these pages are the fallback destination when auth is broken.

These pages are low traffic, so the extra DB hit per render is negligible.

Follow-ups (separate PRs): revoke sessions on ban, replace the client-side auto sign-out redirect with a toast, and make the `INVALID_SESSION` error boundary a static page.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved session validation during sign-in redirects.
  * Anonymous, banned, or revoked sessions are no longer incorrectly redirected to the home page.
  * Added safer handling when session information cannot be retrieved.

### DIFF
--- a/apps/web/src/lib/auth.ts
+++ b/apps/web/src/lib/auth.ts
@@ -552,8 +552,19 @@ export const getUserIdIfLoggedIn = async () => {
 };
 
 export const redirectIfLoggedIn = async () => {
-  const session = await getSession();
-  if (session?.user && !session.user.isGuest) {
+  // Bypass the cookie cache so a revoked or banned session can't bounce the
+  // user back to "/" (RAL-1313).
+  const session = await authLib.api
+    .getSession({
+      headers: await headers(),
+      query: { disableCookieCache: true },
+    })
+    .catch((error) => {
+      logger.error({ error }, "Failed to get session");
+      return null;
+    });
+
+  if (session?.user && !session.user.isAnonymous && !session.user.banned) {
     redirect("/");
   }
 };


lukevella:
Closing unmerged — this approach doesn't hold up:

1. **It isn't actually authoritative in production.** Better-auth is configured with `secondaryStorage` (Redis), and when secondary storage is enabled sessions live only there, stored as a `{session, user}` snapshot written when the session is created. `disableCookieCache: true` reads that Redis snapshot, not Postgres — so a direct Prisma write like `banned: true` on the user row is just as invisible to this check as it was to the cookie cache. The change adds a Redis round-trip per auth-page render and gains nothing.

2. **It reverses a deliberate trust model.** Reads and navigation decisions trust the session cookie cache; validation against the source of truth happens when the user fetches sensitive data or performs a mutation (`mutationSessionGuard` / `authActionClient`), which revoke the session if the user is gone. Splitting that — making `/login` check a fresher source than renders do — is the same asymmetry that caused redirect loops previously.

The actual fix for RAL-1313 is elsewhere: revoke sessions through better-auth's API when banning (direct `prisma.session.deleteMany` is a no-op in cloud prod since sessions aren't in the Session table there), and remove the automated sign-out+redirect legs on the client so a disagreement can't loop. Worst case after that: one bounce into a static "session no longer valid" page, self-healing within the 5-minute cookie cache window.

PR #2599, opened 2026-07-10
🔒 Hide deleted poll data from unauthenticated callers

lukevella:
## Summary

A soft deleted poll stayed fully readable through the public tRPC read endpoints. `polls.get`, `polls.participants.list`, and `polls.comments.list` each looked the poll up by id and returned its data **without checking the `deleted` flag**, so an unauthenticated caller who knew (or still held) a poll id could read the deleted poll's title, description, options, owner identity, participant names/emails, votes, comments, and invite link.

Confirmed against the reported endpoint:

```
/api/trpc/polls.get,polls.participants.list,polls.comments.list
```

Poll ids are 12 char base62 nanoids, so this is **not** an enumeration/mass-harvesting vector — the realistic exposure is anyone who already holds a poll id or invite link (a former participant, a forwarded email) continuing to read the data after the owner deleted the poll. Still a real "delete doesn't revoke access" defect worth fixing.

## Fix

Gate deleted-poll data behind manage access in all three public reads:

- **`polls.get`** — reuses the already-computed `canManage`; deleted poll + non-manager → `NOT_FOUND`. Managers still receive it so the management UI can render the "deleted" screen and restore the poll.
- **`polls.participants.list`** — deleted poll + non-manager → `NOT_FOUND`.
- **`polls.comments.list`** — deleted poll + non-manager → `[]` (mirrors how the endpoint already behaves for a poll with no comments, so a deleted poll is indistinguishable from a non-existent one).

To a non-manager a deleted poll now looks the same as a poll that never existed.

## Test

`apps/web/tests/poll-deleted-access.spec.ts` reproduces the attack end to end: it seeds a soft deleted poll with a participant email and a comment, hits all three endpoints unauthenticated, and asserts the sensitive data is gone (404 / empty), plus a positive control that a live poll is still served. Verified failing before the change and passing after.

## Summary by CodeRabbit

* **Bug Fixes**
  * Deleted polls are now handled consistently across poll details, including returning a “not found” error instead of continuing with the normal payload.
  * Deleted polls no longer return comments; comment lists now return empty results.
  * Deleted polls no longer return participants; participant requests now return “not found.”
  * Poll context UI no longer includes special “deleted poll” rendering, simplifying the displayed poll state.


apps/web/src/trpc/routers/polls.ts
```diff
@@ -684,6 +684,16 @@ export const polls = router({
...
 
+      // A deleted poll must only be visible to a user who can manage it, so
+      // they can still see the "deleted" state and restore it. To everyone
+      // else it is indistinguishable from a poll that never existed.
+      if (res.deleted && !canManage) {
+        throw new TRPCError({
+          code: "NOT_FOUND",
+          message: "Poll not found",
+        });
+      }
```
lukevella: Revert this. Deleted polls shouldn't visible to anyone.

lukevella: Done in a9127cb — the \`!canManage\` exception is gone, so \`polls.get\` returns NOT_FOUND for a deleted poll for everyone including its owner and space managers. Also removed the now-unreachable "deleted poll" screen in poll-context.tsx.

apps/web/src/trpc/routers/polls.ts
```diff
@@ -684,6 +684,15 @@ export const polls = router({
         ? await canUserManagePoll(ctx.user, res)
         : false;
 
+      // A deleted poll is treated as if it never existed, for everyone
+      // including its owner and space managers.
+      if (res.deleted) {
```
lukevella: Move the check further up so we have `!res || res.deleted`

apps/web/src/trpc/routers/polls/participants.ts
```diff
@@ -121,13 +121,19 @@ export const participants = router({
...
+          deleted: true,
         },
       });
 
       if (!poll) {
         throw new TRPCError({ code: "NOT_FOUND", message: "Poll not found" });
       }
 
+      // A deleted poll never exposes its participants.
+      if (poll.deleted) {
```
lukevella: Combine the two if statements in to one

apps/web/src/trpc/routers/polls/comments.ts
```diff
@@ -35,9 +35,15 @@ export const comments = router({
         select: {
           userId: true,
           hideParticipants: true,
+          deleted: true,
         },
       });
 
+      // A deleted poll never exposes its comments.
+      if (poll?.deleted) {
```
lukevella: Similar to participants which should throw a `NOT_FOUND` if `!poll || poll.deleted`

apps/web/src/trpc/routers/polls.ts
lukevella: Done in c069fcb58 — polls.get now does `if (!res || res.deleted)` before computing canManage.

apps/web/src/trpc/routers/polls/participants.ts
lukevella: Done in c069fcb58 — Merged into `if (!poll || poll.deleted)`.

apps/web/src/trpc/routers/polls/comments.ts
lukevella: Done in c069fcb58 — Now throws NOT_FOUND on `!poll || poll.deleted`, matching participants; also dropped the optional chaining on poll afterwards.

PR #2632, opened 2026-07-12
🐛 Fix admin-setup redirect loop and authorize admin access against the database

lukevella:
## Description

Fixes the redirect loop on `/admin-setup` in production: clicking "Make me an admin" updated the role directly in the database, but Better-Auth's session cookie cache (5-minute `maxAge`) still held `role: "user"`, so `/control-panel` bounced the user straight back to `/admin-setup` until the cookie expired.

Two changes:

1. **Refresh the session cookie after promotion** — `makeMeAdminAction` now calls `auth.api.getSession({ query: { disableCookieCache: true } })` after updating the role, re-issuing the cookie from the database before redirecting. This keeps the client-side session (e.g. the control panel menu item) consistent immediately.

2. **Authorize admin access against the database, never the cookie** — the session's role claim can be stale for up to the cookie cache's `maxAge`, so the admin gates no longer trust it:
   - `adminProcedure` (tRPC) now fetches the role from the database
   - `createAdminSSRHelper` now uses `getCurrentUser()` (DB-backed, also handles banned/guest users)
   - the `/admin-setup` page reads the role via `getCurrentUser()`
   - `adminActionClient` already checked the database via `getCurrentUser()` — unchanged

The integration test is also hardened: it previously only waited for the `/control-panel` URL — which the transient redirect satisfies even when the loop occurs — and asserted the role in the database, which was always correct. It now asserts the control panel actually renders and the URL stays on `/control-panel`. Verified locally that the hardened test fails on the old action code (reproducing the loop) and that the admin-setup and control-panel suites pass with these changes.

## Checklist

https://claude.ai/code/session_0172b7m13iWBM98N2o3My8RS

---
_Generated by [Claude Code](https://claude.ai/code/session_0172b7m13iWBM98N2o3My8RS)_

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved the initial admin setup flow to reliably recognize newly granted administrator access.
  - Ensured users are redirected to the control panel with an up-to-date session after becoming an administrator.
  - Strengthened administrator access checks by verifying the current role before allowing access.
  - Improved handling of unauthenticated or restricted users, including safe redirection to login or appropriate access pages.
- **Tests**
  - Added verification that the control panel fully loads after completing admin setup.

### DIFF
--- a/apps/web/src/app/[locale]/admin-setup/actions.ts
+++ b/apps/web/src/app/[locale]/admin-setup/actions.ts
@@ -1,8 +1,10 @@
 "use server";
 
 import { prisma } from "@rallly/database";
+import { headers } from "next/headers";
 import { redirect } from "next/navigation";
 import { isInitialAdmin } from "@/features/setup/utils";
+import { authLib } from "@/lib/auth";
 import { AppError } from "@/lib/errors/app-error";
 import { authActionClient } from "@/lib/safe-action/server";
 
@@ -22,6 +24,16 @@ export const makeMeAdminAction = authActionClient
       },
       data: {
         role: "admin",
+      },
+    });
+
+    // The session cookie cache still holds the old role, which would send
+    // /control-panel straight back here. Re-issue it from the database
+    // before redirecting.
+    await authLib.api.getSession({
+      headers: await headers(),
+      query: {
+        disableCookieCache: true,
       },
     });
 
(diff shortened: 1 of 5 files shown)


apps/web/src/app/[locale]/admin-setup/actions.ts
```diff
@@ -25,5 +27,15 @@ export const makeMeAdminAction = authActionClient
       },
     });
 
+    // The session cookie cache still holds the old role, which would send
+    // /control-panel straight back here. Re-issue it from the database
+    // before redirecting.
+    await authLib.api.getSession({
+      headers: await headers(),
+      query: {
+        disableCookieCache: true,
```
Comment (lukevella) [inline apps/web/src/app/[locale]/admin-setup/actions.ts:36]: @coderabbitai Is this meant to refresh the cookie cache?

PR #2634, opened 2026-07-12
✨ Add SMTP_DEBUG option for verbose SMTP logging

### LINKED ISSUE #1952: Nodemailer verbose logging
Please enable setting verbose logs on nodemailer so we can troubleshoot mail issues:
I am currently struggling to understand why rallly has a problem sending emails using my smtp.

```js
  const transporter = nodemailer.createTransport({
    host: process.env.MAIL_HOST,
    port: Number(process.env.MAIL_PORT),
    secure: process.env.SMTP_SECURE === 'true', // true for 465, false for other ports
    requireTLS: process.env.SMTP_REQUIRE_TLS === 'true',
    auth: {
      user: process.env.MAIL_USER,
      pass: process.env.MAIL_PASS,
    },
    tls: {
      rejectUnauthorized: process.env.SMTP_REJECT_UNAUTHORIZED !== 'false', // allow self-signed
    },
    debug: true, // show debug output
    logger: true, // log to console
  });
```

Image

lukevella:
Fixes #1952

## What

Adds an opt-in `SMTP_DEBUG` environment variable. When set to `true`, the full SMTP conversation (connection, handshake, protocol exchange) is logged through the shared pino logger, giving self-hosters actionable output when diagnosing email delivery issues.

## How

- `packages/emails/src/transport.ts`: passes nodemailer's `debug`/`logger` options when `SMTP_DEBUG=true`. A small adapter maps pino to nodemailer's bunyan style logger interface (pino exposes `level` as a property while nodemailer's type expects a method). The child logger runs at trace level so transcript lines are emitted regardless of the global `LOG_LEVEL`.
- `apps/web/src/env.ts`: declares the new variable alongside the other SMTP flags.
- `apps/docs/self-hosting/configuration.mdx`: documents the variable, marked as available from v4.12.0, with a warning that debug output includes the auth exchange (credentials are only base64 encoded), so it should only be enabled while troubleshooting.

Off by default. No behavior change unless the flag is set.

## Verification

Sent mail through the transport against local mailpit:
- `SMTP_DEBUG=true`: full SMTP transcript emitted as structured JSON under `name: "smtp"` (DNS resolution, EHLO exchange, handshake, message send)
- unset: no transcript output, mail sends as before

`pnpm --filter @rallly/emails type-check` and `pnpm --filter @rallly/web type-check` pass.

## Summary by CodeRabbit

* **New Features**
  * Added optional SMTP conversation debug logging to assist with diagnosing email delivery issues.
  * Enabled via the `SMTP_DEBUG` environment setting.
* **Bug Fixes**
  * Improved compatibility between application logging and email transport diagnostics.
* **Documentation**
  * Documented the new `SMTP_DEBUG` option (default: `false`) in the self-hosting email (SMTP) configuration guide.

### DIFF
--- a/packages/emails/src/transport.ts
+++ b/packages/emails/src/transport.ts
@@ -1,9 +1,27 @@
 import * as aws from "@aws-sdk/client-ses";
 import { defaultProvider } from "@aws-sdk/credential-provider-node";
+import { logger } from "@rallly/logger";
 import type { Transporter } from "nodemailer";
 import { createTransport } from "nodemailer";
+import type { Logger as NodemailerLogger } from "nodemailer/lib/shared";
 
 type EmailProvider = "ses" | "smtp";
+
+// Adapts pino to nodemailer's bunyan-style logger interface. Pino exposes
+// `level` as a property while nodemailer's type expects a method (it never
+// calls it).
+function createSmtpDebugLogger(): NodemailerLogger {
+  const smtpLogger = logger.child({ name: "smtp" }, { level: "trace" });
+  return {
+    level: () => {},
+    trace: smtpLogger.trace.bind(smtpLogger),
+    debug: smtpLogger.debug.bind(smtpLogger),
+    info: smtpLogger.info.bind(smtpLogger),
+    warn: smtpLogger.warn.bind(smtpLogger),
+    error: smtpLogger.error.bind(smtpLogger),
+    fatal: smtpLogger.fatal.bind(smtpLogger),
+  };
+}
 
 export type SupportedEmailProviders = EmailProvider;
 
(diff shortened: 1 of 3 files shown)


packages/emails/src/transport.ts
```diff
@@ -40,6 +58,11 @@ function createTransportForProvider(provider: EmailProvider): Transporter {
       const rejectUnauthorized =
         process.env.SMTP_REJECT_UNAUTHORIZED !== "false";
 
+      // Logs the full SMTP conversation including the auth exchange, where
+      // credentials are only base64 encoded. Never enable outside of
+      // troubleshooting.
+      const debug = process.env.SMTP_DEBUG === "true";
```
lukevella, line 64: @claude we should be using env.SMTP_DEBUG here.

PR #2635, opened 2026-07-12
📝 Document SMTP_DEBUG environment variable

lukevella:
Documents the `SMTP_DEBUG` variable added in #2634.

Kept separate from the feature PR because docs deploy from main immediately. Merge this when v4.12.0 is released so the docs match what self-hosters can actually use.


lukevella:
Superseded by the doc-updates branch, which accumulates docs changes to merge as part of the release workflow.

PR #2638, opened 2026-07-12
🐛 Authorize admin mutations against the database, not the session snapshot

lukevella:
## Problem

Banning a user in prod failed with `APIError: You are not allowed to ban users` even though the caller is an admin.

With `secondaryStorage` enabled, better-auth serializes `{ session, user }` into Redis at session creation. `findSession` returns that snapshot without touching the database, and the daily session refresh carries the old user object forward. Since admin roles were granted with raw Prisma writes (`makeMeAdminAction`, `changeRoleAction`), better-auth never refreshed the snapshots — the database said `admin` while every active session still said `user`. Rallly's own checks pass (they authorize against the database since #2632), but `authLib.api.banUser` runs better-auth's internal check against the stale session role and throws.

The error was also invisible: `handleServerError` captured it in Sentry and collapsed it to a generic `INTERNAL_SERVER_ERROR` toast.

## Changes

- **Role changes go through `internalAdapter.updateUser`** (new `updateUserRole` mutation) so `refreshUserSessions` rewrites the Redis snapshots. This also makes the `disableCookieCache` re-issue in admin-setup actually effective with secondary storage.
- **Ban/unban go through `internalAdapter`** instead of the admin plugin endpoints, which authorize against the session snapshot. Authorization stays where it already is: CASL against the database (including the self-ban guard). `banUserAsSystem` is merged into `banUser` — one code path for admin bans and the moderation auto-ban.
- **Better-auth `APIError` codes surface through safe-action** so future failures toast the real error code instead of a generic internal server error. Sentry capture is kept.

## Verification

Drove the dev app end-to-end (Redis secondary storage active): logged in as the seeded control panel admin, banned `dev@rallly.co` via the UI (dialog closed, DB shows `banned=t`, reason, `bannedAt`), then unbanned (flags cleared). Type check and Biome pass.

## Note for prod

Existing stale sessions are not retroactively fixed — re-login once after deploy so your session snapshot picks up the admin role.

Supersedes #2637, which was accidentally branched off a WIP base.

## Summary by CodeRabbit

* **Improvements**
  * Admin role changes now follow a consistent authorization flow and refresh access promptly.
  * Banned-domain moderation enforcement continues to flag content and restrict affected accounts.
  * Account bans now revoke active sessions, while unbanning properly restores account access.
* **Bug Fixes**
  * Improved handling of authentication and service errors so users receive more accurate status messages, including permission, rate-limit, and availability errors.

### DIFF
--- a/apps/web/src/app/[locale]/admin-setup/actions.ts
+++ b/apps/web/src/app/[locale]/admin-setup/actions.ts
@@ -1,9 +1,9 @@
 "use server";
 
-import { prisma } from "@rallly/database";
 import { headers } from "next/headers";
 import { redirect } from "next/navigation";
 import { isInitialAdmin } from "@/features/setup/utils";
+import { updateUserRole } from "@/features/user/mutations";
 import { authLib } from "@/lib/auth";
 import { AppError } from "@/lib/errors/app-error";
 import { authActionClient } from "@/lib/safe-action/server";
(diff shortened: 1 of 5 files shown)


lukevella:
@coderabbitai review

PR #2643, opened 2026-07-13
✨ Cap API keys per space and migrate writes to server actions (RAL-1268)

lukevella:
Closes RAL-1268.

## Summary

Migrates the API key write endpoints off tRPC to server actions (per the queries-only tRPC rule), then layers the launch-hygiene items on top of the new mutation layer.

## Migration (tRPC → server actions)

- New `features/api-keys/{schema,ability,mutations,actions}.ts`.
- `apiKeys.create` / `apiKeys.revoke` removed from the tRPC router; `apiKeys.list` stays (it's a query).
- The `isApiAccessEnabled` gate moves into the actions.
- The revoke procedure's hand-rolled `spaceId` ownership check is replaced by a CASL ability (`revoke ApiKey` scoped to the acting space).
- Both button components call the actions and invalidate the tRPC `list` query on success.

## Hygiene

- **Cap:** active (non-revoked, non-expired) keys capped at 10 per space, enforced in the create mutation. Clutter prevention, not a security control — check-then-create is intentionally non-transactional.
- **PostHog:** `developer:api_key_create` / `developer:api_key_revoke` for launch adoption.

## Notes

- Cap breaches surface as `TOO_MANY_REQUESTS`; the create dialog shows a cap-specific message (`apiKeyLimitReached`) rather than the generic rate-limit toast.
- "Already revoked" now throws `FORBIDDEN` (was tRPC `BAD_REQUEST`, which isn't in our `AppErrorCode` set).

## Verification

type-check, Biome, i18n scan, and the full web unit suite (230 tests) pass.

## Summary by CodeRabbit

* **New Features**
  * API key creation and revocation now use improved secure actions.
  * Added a limit of 10 active API keys per space, with a clear notification when reached.
  * API key settings now load more reliably and display active, revoked, and usage information.
* **Bug Fixes**
  * Improved authorization checks for viewing and revoking keys.
  * Prevented duplicate or already-revoked keys from being revoked again.
* **UI Updates**
  * Simplified API key status and error messaging.


lukevella:
Addressed the CodeRabbit review findings in `8d631d95a`:

- **Duplicate create schema** (nitpick): the create button now imports the shared `createApiKeySchema` from `@/features/api-keys/schema` instead of a local copy, so client and server validation can't drift. Dropped the now-unused `zod` import.
- **Revoke race** (major): the final write is now a conditional `updateMany({ where: { id, revokedAt: null } })`; a 0-row result throws `CONFLICT`. Concurrent revokes can no longer both report success or emit duplicate `developer:api_key_revoke` events.
- **Cap atomicity** (major): **intentionally not changed.** Per RAL-1268 the 10-key cap is clutter prevention, not a security control — the worst case of the check-then-create race is a single extra row, and there is per-space rate limiting upstream. A serializable transaction or per-space advisory lock on every key creation isn't warranted for that.

apps/web/src/lib/safe-action/client.ts
```diff
@@ -29,6 +29,12 @@ export const useSafeAction: typeof useAction = (action, options) => {
               defaultValue: "The resource was not found",
             });
             break;
+          case "CONFLICT":
+            translatedDescription = t("actionErrorConflict", {
+              defaultValue:
+                "This action conflicts with the current state. Please refresh and try again.",
+            });
+            break;
```
lukevella: @claude Remove this. The message is not useful and we don't need to handle it in the global error handler

apps/web/src/features/api-keys/data.ts
```diff
@@ -1,16 +1,73 @@
...
 import { isSelfHosted } from "@/lib/constants";
 import { posthog } from "@/lib/posthog";
 
+/**
+ * Resolve the current user's space and verify API access. Auth is enforced
+ * here in the DAL so no call site can fetch API keys without the gate.
+ * Failures are returned as values; stale sessions still throw
+ * InvalidSessionError from getCurrentUser for the boundary to handle.
+ */
+export async function getApiKeyAccessContext() {
```
lukevella: Use react cache to avoid running this multiple times in one render

PR #2651, opened 2026-07-13
✨ Move to a per-space API rate limit (RAL-1262)

lukevella:
## Summary

Replaces the five per-endpoint rate limiters (each keyed on `apiKeyId`) with a single limit keyed on `private-api:{spaceId}`. Previously, creating more API keys multiplied a space's throughput; now the limit is per space and easy to publish.

**Limit: 60 requests/minute per space, uniform across all endpoints.** Starting conservative — easier to raise later than to lower a limit integrations already depend on.

## Changes

- New `rateLimit` middleware in `apps/web/src/app/api/private/utils/rate-limit.ts`, registered after `spaceApiKeyAuth` on every poll route
- Dropped the five per-endpoint `rateLimiter` instances from `route.ts`
- 429 body matches `errorResponseSchema` with code `RATE_LIMIT_EXCEEDED`
- `hono-rate-limiter` emits the standard `RateLimit-*` headers and a `Retry-After` header on 429
- OpenAPI spec documents the limit: a rate-limits section in the API description plus a `429` response on each endpoint

## Known limitation (noted in code)

When Redis/KV is unavailable the fallback is a per-instance memory store, which on Vercel is effectively no limit. Accepted for now; request logging makes it observable.

## Testing

- `pnpm test:unit` — added tests asserting `RateLimit-*` headers on success and a `429` with `RATE_LIMIT_EXCEEDED` + `Retry-After` once the per-space limit is exceeded (66 pass)
- `pnpm type-check` and `pnpm check` clean
- Verified the generated OpenAPI spec renders the rate-limits description and the `429` response on all five endpoints

Closes RAL-1262

## Summary by CodeRabbit

* **New Features**
  * Added centralized per-space rate limiting to the private Polls API endpoints (60 requests/min).

* **Bug Fixes**
  * Rate-limited responses now include standard `RateLimit-Limit` and `RateLimit-Remaining` headers.
  * When the limit is exceeded, requests return **HTTP 429** with `RATE_LIMIT_EXCEEDED` and a `Retry-After` header.

* **Documentation**
  * Updated OpenAPI documentation to clearly describe the shared per-space rate-limit behavior and `Retry-After`.

* **Tests**
  * Added tests covering `RateLimit-*` headers and `429` handling when the limit is exceeded.

### DIFF
--- a/apps/web/src/app/api/private/[...route]/route.test.ts
+++ b/apps/web/src/app/api/private/[...route]/route.test.ts
@@ -1106,6 +1106,63 @@ describe("Private API - /polls", () => {
     });
   });
 
+  describe("Rate limiting", () => {
+    // The rate limiter is keyed per space and shares a module-level memory
+    // store across the whole file. Use a dedicated space id here so the
+    // requests below don't count against (or get counted by) other tests.
+    const rateLimitSpaceId = "rate-limit-space-id";
+
+    beforeEach(() => {
+      vi.mocked(prisma.spaceApiKey.findMany).mockResolvedValue([
+        {
+          ...mockApiKey,
+          hashedKey: hashApiKey(testApiKey),
+          spaceId: rateLimitSpaceId,
+          lastUsedAt: new Date(),
+        },
+      ]);
+    });
+
+    it("should include standard RateLimit headers on successful responses", async () => {
+      vi.mocked(prisma.poll.findFirst).mockResolvedValue(null);
+
+      const res = await app.request("/api/private/polls/some-poll", {
+        method: "GET",
+        headers: { Authorization: `Bearer ${testApiKey}` },
+      });
+
+      expect(res.headers.get("RateLimit-Limit")).toBe("120");
+      expect(res.headers.get("RateLimit-Remaining")).not.toBeNull();
+    });
+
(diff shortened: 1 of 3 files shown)


lukevella:
Addressed the CodeRabbit feedback:

**Nitpick — use `RATE_LIMIT_PER_MINUTE` in the tests instead of hardcoded `120`/`121`:** Done in `190f93d`. The header assertion now compares against `String(RATE_LIMIT_PER_MINUTE)` and the loop runs `RATE_LIMIT_PER_MINUTE + 1` times, so the tests stay in sync with the constant.

**Major — memory-store fallback can be bypassed across instances in production:** Not changing this. It's a deliberate, documented decision from the issue (RAL-1262): _"when Redis/KV is unavailable the fallback is a per instance memory store, which is effectively no limit on Vercel. Accepted for now; request logging makes it observable."_ On Rallly Cloud, KV is always configured, so the fallback only takes effect if KV is genuinely down — in which case failing closed (rejecting all API traffic) is worse than degrading the limit. Self-hosters without Redis are single-instance, where the memory store is a correct per-space limiter. The limitation is called out in a comment in `rate-limit.ts`.

PR #2733, opened 2026-07-21
♻️ Migrate space member mutations to server actions

lukevella:
Fixes RAL-1403, RAL-1402

## What

Migrates the five member/invite mutations off the `spaces` tRPC router into server actions, per the writes-are-server-actions rule:

- `spaces.inviteMember` → `inviteMemberAction`
- `spaces.cancelInvite` → `cancelInviteAction`
- `spaces.removeMember` → `removeMemberAction`
- `spaces.changeMemberRole` → `changeMemberRoleAction`
- `spaces.acceptInvite` → `acceptInviteAction` (its own entry point — authorized by the invite row matching the signed-in user's email, not by space membership)

## Shape

- `features/space/member/actions.ts` owns authentication and authorization (safe-action + CASL against database state). Expected outcomes come back as `{ ok, ... }` unions (`ALREADY_MEMBER`, `INVITE_PENDING`, `NOT_ENOUGH_SEATS`, `INVITE_NOT_FOUND`); thrown `AppError`s hit the global `useSafeAction` toast.
- `features/space/member/mutations.ts` trusts its input, takes explicit parameters, and owns the writes. The seat availability check moved in here from the router (returned as a value instead of thrown).
- Client components swap `useMutation` for `useSafeAction`; the global `router.refresh()` keeps the still-tRPC `listMembers`/`listInvites` queries fresh through the page prefetch + `HydrationBoundary`, so the seat count now updates in place — the reload workaround in the remove-member e2e test is gone.
- The accept invite button handles its result from the action promise rather than the hook's `onSuccess`: accepting deletes the invite, so the action response re-renders the route as a 404 and unmounts the button before effect-driven callbacks run.

## RAL-1402 rides along

- `getTotalSeatsForSpace` moves from `features/space/utils.ts` to `features/space/data.ts` — still the single parameterized seat policy function (license tiers, subscription quantity, caps).
- New session scoped, React cached `getSeatUsage()` calls `getActiveSpace()` internally and returns `{ total, used }`; the members and billing settings pages read through it instead of passing plain space ids to the seat functions.
- `features/space/utils.ts` is back to pure helpers (role mapping only).

## Verification

- `pnpm type-check`, `pnpm check`, `pnpm test:unit` pass.
- All 5 `space-members.spec.ts` e2e tests pass, with the `page.reload()` workaround removed from "admin can remove a member".

## Summary by CodeRabbit

* **New Features**
  * Added clearer handling for inviting, accepting, canceling, removing, and updating member roles.
  * Added messages for missing invites and unavailable seats.
  * Seat usage and limits are now reflected consistently across billing and member settings.
* **Bug Fixes**
  * Improved invite and member-management feedback, including loading states and success/error notifications.
  * Prevented invitations and invite acceptance when no seats are available.
* **Tests**
  * Updated member-removal coverage to verify seat counts immediately after removal.


lukevella:
CI's integration failure on "admin can change a member's role" was a race in next-safe-action 8.0.11, not the migration itself: `execute()` defers `setIsExecuting(true)` into a `setTimeout(0)` while the action starts immediately inside a transition. The action POST completed in ~40ms on the CI runner and the starved main thread ran the timer callback *after* the `finally` that clears `isExecuting`, wedging the hook status at "executing" — so `onSuccess` (toast + global `router.refresh()`) never fired even though the role change persisted (verified from the Playwright trace: the action response contained the updated member list). 8.6.0 sets the state synchronously with request-id guards, which is exactly the upstream fix for this shape of bug. Upgraded in 634773e; full suite passes locally against a production build with `CI=1`.

PR #2737, opened 2026-07-21
♻️ Introduce loaders.ts: split request facing reads from the parameterized DAL

lukevella:
Closes RAL-1407

Adopts the loader/action pair to complete the read/write symmetry: request facing reads (session, redirects, page semantics) live in `loaders.ts`, the parameterized core stays in `data.ts`. No behavior changes — pure relocation plus lint and docs.

## Sweep

- **`features/user`**: `getCurrentUser` moves to `loaders.ts`. One deviation from the issue: `createUserDTO` stays in `data.ts` rather than moving — `getCurrentUser` previously queried prisma directly, and loaders can never import `@rallly/database`, so `data.ts` gains a parameterized `getUser(userId)` that owns the query + DTO mapping and `getCurrentUser` delegates to it. That keeps "loaders consume data.ts, never the reverse" intact and mirrors the space side where `createSpaceDTO` stays in `data.ts` (its own queries need it).
- **`features/space`**: `getActiveSpace` and `getSeatUsage` move to `loaders.ts`; `data.ts` no longer imports `next/navigation` or session state.
- **`features/api-keys`**: `data.ts` dissolves to its parameterized core (`getSpaceApiKeys({ spaceId: AuthorizedSpaceId })`, `isApiAccessEnabled`). The page state reads (`getApiKeysPageState`, the old `getSpaceApiKeys` wrapper) had one consumer segment, so `settings/api-keys/page.tsx` now composes the gate itself; `getApiKeyAccessContext` is inlined into `requireApiAccess` in `actions.ts` with identical error mapping.
- Verified no other `features/**/data.ts` reads session state.

## Lint

- The enumerated session scoped module ban in the API route overrides (`@/features/user/data`, `@/features/api-keys/data`) becomes the glob `@/features/**/loaders`; API routes can now import the parameterized `@/features/user/data`.
- `features/**/data.ts` and `mutations.ts` are banned from `next/headers` and `next/navigation`.
- Loaders are absent from the `@rallly/database` allowlist, so database imports are banned there by default. All three rules verified to fire.
- `loaders` added to the feature filenaming convention.

## Docs

CLAUDE.md gains the `loaders.ts` vocabulary entry with the dependency rules, the read/write symmetry table, the route-by-default placement rule with its promotion trigger, and the note that `session-data` naming and `queries.ts` were rejected.

Verified with `pnpm check`, `pnpm type-check`, and `pnpm test:unit` (261 passing).

## Summary by CodeRabbit

* **New Features**
  * Added centralized server-side loaders for the signed-in user, active space, and seat usage, with consistent redirect and error handling.
  * API key settings now gates access per user and space, showing the upgrade UI when access is disabled.
* **Refactor**
  * Updated space- and user-related pages to use the new loading layer (including API keys, billing, members, preferences, profile, and admin setup).
  * API keys data access is now direct per-space and type handling is simplified.
* **Documentation**
  * Expanded architecture guidance and vocabulary/import rules to clearly separate loader behavior from parameterized data access.
* **Chores**
  * Updated linting/feature-structure checks to recognize and restrict loader modules appropriately.


apps/web/src/app/[locale]/(space)/settings/api-keys/page.tsx
```diff
@@ -24,9 +27,10 @@ import { ApiUsageLimits } from "./components/api-usage-limits";
 import { CreateApiKeyButton } from "./components/create-api-key-button";
 
 export default async function ApiKeysSettingsPage() {
-  const pageState = await getApiKeysPageState();
+  const user = await getCurrentUser();
+  const space = user ? await getActiveSpaceForUser(user.id) : null;
```
Comment (lukevella) [inline apps/web/src/app/[locale]/(space)/settings/api-keys/page.tsx]: Shouldn't we be using getActiveSpace() here?

apps/web/src/app/[locale]/(space)/settings/api-keys/page.tsx
```diff
@@ -35,7 +39,11 @@ export default async function ApiKeysSettingsPage() {
     );
   }
```
Comment (lukevella) [inline apps/web/src/app/[locale]/(space)/settings/api-keys/page.tsx]: Shouldn't be necessary. Handled by loaders

apps/web/src/app/[locale]/(space)/settings/api-keys/page.tsx
```diff
@@ -51,36 +59,28 @@ export default async function ApiKeysSettingsPage() {
...
 }
 
-async function ApiKeysContent() {
-  const result = await getSpaceApiKeys();
-
-  if (!result.ok) {
-    notFound();
-  }
+async function ApiKeysContent({ spaceId }: { spaceId: AuthorizedSpaceId }) {
+  const apiKeys = await getSpaceApiKeys({ spaceId });
```
Comment (lukevella) [inline apps/web/src/app/[locale]/(space)/settings/api-keys/page.tsx]: We should be using a loader here rather than passing Around an AuthorizedSpaceId

apps/web/src/app/[locale]/(space)/settings/api-keys/page.tsx
Comment (lukevella) [reply apps/web/src/app/[locale]/(space)/settings/api-keys/page.tsx]: Done in b27cc67e7 — getActiveSpace() replaces the getActiveSpaceForUser composition.

apps/web/src/app/[locale]/(space)/settings/api-keys/page.tsx
Comment (lukevella) [reply apps/web/src/app/[locale]/(space)/settings/api-keys/page.tsx]: Space gating dropped — getActiveSpace owns the /login and /setup redirects now. Kept the if (!user) guard after getCurrentUser since its contract is to return null and let the caller navigate (same as the billing and profile pages), and TS needs the narrowing for isApiAccessEnabled(user, space).

apps/web/src/app/[locale]/(space)/settings/api-keys/page.tsx
Comment (lukevella) [reply apps/web/src/app/[locale]/(space)/settings/api-keys/page.tsx]: Done — ApiKeysContent calls getActiveSpace() itself; React cache means no extra query.

PR #2789, opened 2026-07-30
✨ Add participant response note, visible to the host only

lukevella:
Fixes [RAL-1469](https://linear.app/rallly/issue/RAL-1469/add-participant-response-note-visible-to-the-host-only)

Part 1 of the discussions swap: an optional note submitted with a response, seen only by the host.

## Changes

- **Data**: nullable `note` text column on `participants`; `MAX_RESPONSE_NOTE_LENGTH = 500` next to the comment constants
- **Response form**: optional "Note for the organizer" textarea in the new participant dialog, wired to `MaxCharLength` with the 500 cap and a persistent visibility disclosure ("Only the organizer will see this note")
- **Server side enforcement**: `participants.list` now resolves poll admin access once and strips `note` from every non admin payload, so other participants never receive notes at the API level. Input is trimmed, capped at 500, and whitespace only notes collapse to null (`responseNoteInput`, unit tested)
- **Host display**: note icon on the participant row (desktop table and mobile header) opening a popover with the note, linkified and pre wrap, rendered only on the admin view
- **Email**: the new participant email quotes the note in a bordered block (same pattern as the new comment email), uses the note as preview text, and sets `Reply-To` to the participant's email when known via a new `replyTo` passthrough in `sendRenderedEmail`. Reply guidance ("You can reply to this email to respond to {name}") appears only when both a note and a reply address exist
- **Instrumentation**: `has_note` property on `poll_response_submit` plus a `poll_response:note_submit` event with `note_length`
- **Tests**: added vitest setup to `packages/emails` (mirroring `packages/ui`); render tests cover the quote block, HTML escaping of user content, the no note case, and reply guidance gating. Assertions target default copy and prop content since ICU interpolation does not run under vitest

## Verification

- `pnpm check`, `pnpm type-check`, `pnpm test:unit` all green (only pre existing biome warning in rich-text-editor)
- Payload enforcement is code level in `participants.list`; notes editing, participant facing display, and broadcast remain out of scope per the issue

## Summary by CodeRabbit

* **New Features**
  * Organizers can add optional notes when adding poll participants, with a 500-character limit.
  * Notes are visible only to organizers and the participant who submitted them.
  * Participant notes are displayed through a new popover interface with edit, rename, and delete actions.
  * Notification emails include participant notes and reply guidance when applicable.
* **Bug Fixes**
  * Improved note handling by trimming whitespace and preventing empty notes from being saved.
* **Tests**
  * Added coverage for note validation, email rendering, escaping, and reply guidance.


lukevella:
We're going to want to track usage of this feature in posthog. Update analytics so we can see what the adoption is.

lukevella:
Adoption tracking as of 8ed16436c:

1. **`new_participant_dialog:add_note_button_click`** (client) — fires when the Add a note chip is clicked, with `pollId`, `spaceId`, `tier`, and poll/space groups. Measures discovery and intent.
2. **`poll_response_submit`** (server) — now carries `has_note` and `note_length` on every response. Adoption rate is a single breakdown (`has_note = true`), and `note_length` gives the length distribution for validating the 500 cap.

Funnel: chip clicks (1) vs responses with `has_note` (2) shows drop off after opening the field.

There was briefly a dedicated `poll_response:note_submit` event, but it fired at the same instant as `poll_response_submit` and carried nothing a property could not — removed as redundant.

Not tracked: host side note views (popover opens). That is engagement rather than adoption — say the word if you want it.

PR #2840, opened 2026-08-03
♻️ Rebuild profile settings on Setting primitives

lukevella:
## Description

Brings the profile page in line with the preferences and notifications pages by composing it from the shared `Setting` primitives, continuing the work from #2836 and #2838.

Picture, name and email are now three rows in a single `SettingsGroup`, replacing the stacked form plus a separate email card.

**Email no longer has its own section.** Its two step flow (request a change, then verify a 6 digit code) can't collapse into the Setting control column, so it moved into a dialog opened from the row, following the "pages for browsing, dialogs for actions" convention. `profile-email-address.tsx` is deleted and its logic now lives in `change-email-dialog.tsx`, driven by `useDialog`. Success is reported with a toast instead of an inline alert, since the dialog closes on completion.

`i18n:scan` added the new keys and pruned three that went unused.

### Notes for the reviewer

Two things worth a look:

**The name row keeps an explicit Save button**, which is the one place this deviates from the auto save pattern on preferences. There was no clean slot for it: the `Setting` grid has no button area, and `SettingHint` renders an InfoIcon for explanatory text, so putting a button there would misuse the primitive. The button therefore sits below the row and only appears once the field is dirty. Auto save on blur would remove that wart if we'd rather have consistency.

**This hasn't been checked visually.** Browser verification was blocked in my environment, so I verified server side (route compiles, `GET /settings/profile 200`) and added `profile-settings.test.tsx` covering the three rows rendering, Save appearing only when dirty, and the dialog opening pre-filled. Worth a glance before merging, particularly how the avatar and its "Choose…" button sit in the right hand control column, which is narrower than the full width layout they were designed for.

`ProfilePicture` gains a wrapper element so it can accept the aria attributes `SettingControl` clones onto its child.

## Checklist

## Summary by CodeRabbit

- **New Features**
  - Updated profile settings with separate sections for name, email address, and profile picture.
  - Added a two-step email change flow with one-time passcode verification.
  - Added clearer descriptions and validation messages for profile updates.
- **Accessibility**
  - Improved profile picture support for accessible labeling and descriptions.
- **Bug Fixes**
  - Added handling for duplicate email addresses, invalid or expired verification codes, and failed requests.
- **Tests**
  - Added coverage for profile settings actions and email-change dialog behavior.


lukevella:
Needs more work — the email change flow doesn't look right yet, and the Danger Zone still needs the same treatment.

PR #2866, opened 2026-08-04
♻️ Convert control panel version tile to a server component

lukevella:
## Summary

Moves the control panel's version/update-status tile off tRPC and onto the server. The `system` router had no other procedures, so it is deleted and unregistered (and removed from the biome migration allowlist).

- **`features/instance-settings/service.ts`** (new) — wraps the update API (`API_BASE_URL/api/updates`). Fetches with `next: { revalidate: 3600 }` (preserves the 1-hour cache semantics via the data cache, replacing the KV cache) and a hard `AbortSignal.timeout(3000)`. Any failure — timeout, non-OK, bad payload — resolves to `null` and logs a single structured warn.
- **`features/instance-settings/data.ts`** — `getInstanceSettings` now also returns `instanceId` (same cached singleton query).
- **`version-tile.tsx`** — server component rendered under its own `<Suspense>` in the page, with a fallback that is the same tile linking to the static releases page. The `isFetching` spinner is gone.

## Resilience

The control panel shell streams immediately and is unaffected by an update-API outage. Verified `patch-fetch.js` in Next 16.3: passing an `AbortSignal` does not opt the fetch out of the data cache, and the signal is stripped on background revalidation, so the timeout and `revalidate: 3600` compose. The timeout matters on the cold-cache path (fresh container start during an outage); after one successful check, stale-while-revalidate hides outages entirely.

## Verification

- `pnpm type-check`, `pnpm check`, `pnpm build` all green
- `control-panel.spec.ts`: 3/3 passing
- Outage path: ran the admin flow with `API_BASE_URL` pointed at a blackhole IP (hung connection) — page renders immediately, tile streams in with the static releases link, no update badge, no error UI
- Success path: ran against a local mock update API returning `v99.0.0` — tile streams in with the "Update available" badge linking to the release, confirming the async component resolves and replaces the fallback

## Summary by CodeRabbit

* **New Features**
  * Added instance update-status checking to the control panel.
  * Version information now links directly to the releases page.
  * Update notifications appear when a newer version is available.
  * Added an “Up to date” status when no update is available.
* **Improvements**
  * Added a loading placeholder while update information is retrieved.
  * Improved update handling with validated version comparisons and graceful error handling.


apps/web/src/app/[locale]/control-panel/page.tsx
```diff
@@ -129,7 +130,9 @@ export default async function AdminPage() {
               </TileTitle>
             </Tile>
             {/* VERSION */}
-            <VersionTile />
+            <Suspense fallback={<VersionTileFallback />}>
+              <VersionTile />
```
Comment (lukevella) [inline apps/web/src/app/[locale]/control-panel/page.tsx]: Can we render the version tile and only wrap the contents of the TileDescription in Suspense. We can render a simple Skeleton component while the component is being rendered.

apps/web/src/features/instance-settings/service.ts
```diff
@@ -0,0 +1,62 @@
...
+  try {
+    const res = await fetch(url, {
+      // Node fetch has no default timeout — without this a hung upstream
+      // holds the streamed slot open indefinitely on a cold cache
+      signal: AbortSignal.timeout(3000),
+      next: { revalidate: 3600 },
+    });
+    if (!res.ok) return null;
+    const parsed = updateStatusSchema.safeParse(await res.json());
+    if (!parsed.success) return null;
```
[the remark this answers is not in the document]
lukevella: Added warn logs with instanceId/version/status (and zod issues for parse failures) to both branches in 266081b84.

apps/web/src/features/instance-settings/service.ts
```diff
@@ -0,0 +1,62 @@
...
+  const len = Math.max(a.length, b.length);
+  for (let i = 0; i < len; i++) {
+    const ai = a[i] ?? 0;
+    const bi = b[i] ?? 0;
+    if (ai !== bi) return ai < bi;
+  }
+  return false;
+}
+
+export async function getUpdateStatus({ instanceId }: { instanceId: string }) {
```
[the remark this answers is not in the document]
lukevella: Skipping: repo precedent uses a class/factory only where there's per-instance state to encapsulate (Stripe SDK client in billing, OAuth credentials in calendars). This is a stateless single-endpoint fetch — a factory would be ceremony with no behavior to own.

apps/web/src/app/[locale]/control-panel/version-tile.tsx
```diff
@@ -1,39 +1,28 @@
...
-          target="_blank"
-          rel="noreferrer noopener"
-        />
-      }
-    >
+    <Tile render={<a href={url} target="_blank" rel="noreferrer noopener" />}>
       <PageIcon>
         <DownloadIcon />
       </PageIcon>
       <TileTitle>{`v${appVersion ?? "unknown"}`}</TileTitle>
```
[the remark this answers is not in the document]
Comment (lukevella) [reply apps/web/src/app/[locale]/control-panel/version-tile.tsx:46]: Skipping: the "unknown" fallback is carried over unchanged from the previous client tile and only renders when NEXT_PUBLIC_APP_VERSION is unset, which never happens in a production build (build:web and Docker both inject it). Adding an i18n key would create Crowdin churn for a dev-only string.

apps/web/src/app/[locale]/control-panel/page.tsx
Comment (lukevella) [reply apps/web/src/app/[locale]/control-panel/page.tsx]: Done in 2a27a1bfd — the tile shell renders statically and only the TileDescription content suspends with a Skeleton fallback. One trade-off: the tile link is now always the releases index (the anchor can't depend on the fetch result anymore), where previously an available update deep-linked to the specific release.

apps/web/src/app/[locale]/control-panel/version-tile.tsx
```diff
@@ -1,43 +1,45 @@
...
-  const { data, isFetching } = trpc.system.getUpdateStatus.useQuery(undefined, {
-    staleTime: 60 * 60 * 1000,
-    refetchOnWindowFocus: false,
-  });
+async function UpdateStatus() {
+  const { instanceId } = await getInstanceSettings();
+  const update = instanceId ? await getUpdateStatus({ instanceId }) : null;
+
+  if (!update) {
+    return null;
```
Comment (lukevella) [inline apps/web/src/app/[locale]/control-panel/version-tile.tsx:16]: I think the message should be "Up to date" in green-600

Comment (lukevella) [reply apps/web/src/app/[locale]/control-panel/version-tile.tsx:16]: Done in e347fb5ca — UpdateStatus now resolves the string via getTranslation from @/i18n/server with a defaultValue, removing the client boundary. Verified i18n:scan keeps the existing updateAvailable key untouched.

PR #2968, opened 2026-08-17
💄 Rename CTA experiment variant to create-free-poll

lukevella:
Follow-up to #2963. Luke flagged that "Get started for free" → poll creation wizard is a label to destination mismatch: it promises onboarding, then asks the visitor to configure a poll. The variant also changed two things at once (action specificity + the free claim), so a result would have been unattributable.

New challenger: **"Create a free poll"** — same verb, same object, same destination match as control; the only delta is surfacing *free* at the point of action.

- Variant key `get-started` → `create-free-poll` everywhere (CtaLabel, shell/page prop types, middleware rewrite constant, variant route directory)
- i18n: `getStartedForFree` removed, `createAFreePoll` added (via `pnpm i18n:scan`)
- Stale `get-started` cookies (≤1h TTL) degrade to control

PostHog side already done: [experiment 90945](https://eu.posthog.com/project/1971/experiments/90945) analysis reset to draft (the few hours of exposures are discarded from analysis), flag disabled, variant renamed to `create-free-poll`, metrics preserved, hypothesis updated. Nobody is currently exposed to any variant.

**Launch sequence: merge → wait for the landing deploy → press Launch on the experiment.** Launching before the deploy would serve `create-free-poll` to code that doesn't recognize it (visitors would see control while exposures record the variant).

Verified locally: `create-free-poll` cookie → variant HTML server-rendered at `/` (both CTAs); stale `get-started` cookie → control; no cookie with flag disabled → control cached.

## Summary by CodeRabbit

* **Improvements**
  * Updated the landing page call-to-action wording to “Create a free poll.”
  * Updated the CTA experiment variant to consistently use the new wording across landing page layouts and pages.
  * Replaced the previous “Get started for free” localization with the clearer “Create a free poll” label.

### DIFF
--- a/apps/landing/public/locales/en/home.json
+++ b/apps/landing/public/locales/en/home.json
@@ -1,4 +1,5 @@
 {
+  "createAFreePoll": "Create a free poll",
   "createAPoll": "Create a poll",
   "doodleAlternative": "The best free Doodle alternative",
   "doodleAlternativeDescription": "Rallly is the Doodle alternative that everyone is looking for. Thousands of users have already made the switch and are now enjoying professional ad-free meeting polls in an intuitive and easy-to-use interface.        ",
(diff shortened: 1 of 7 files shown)


lukevella:
Superseded — decision is to not run CTA experiments at all. Teardown PR removing the experiment scaffolding follows.

